PR #1, opened 2016-01-04
base-node-mocha test using best practices

glenna-m:
Modified the base-node-mocha test to use best practice BDD syntax.
Also added a test case for the assign transform path to get 100% coverage on base-node module.

These changes were done to provide a clean "getting started" example that the noflo-rdf-component wiki test documentation (https://github.com/rdf-pipeline/noflo-rdf-components/wiki/Component-Testing) will refer to.

### DIFF
--- a/test/base-node-mocha.js
+++ b/test/base-node-mocha.js
@@ -4,7 +4,6 @@
  */
 
 var chai = require('chai');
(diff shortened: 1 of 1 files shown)


dbooth-boston:
LGTM

PR #4, opened 2016-01-11
Remove group support in round-robin (parallel and nested groups weren…

catch-point:
Remove group support in round-robin (parallel and nested groups weren't supported anyway)


glenna-m:
LGTM.

PR #5, opened 2016-01-13
Promise component

catch-point:
Use promises to write asynchronous  components

### DIFF
--- a/components/http-accept.js
+++ b/components/http-accept.js
@@ -10,16 +10,6 @@ var promiseComponent = require('./promise-component');
 exports.getComponent = promiseComponent({
     description: "Extracts the request body as a String and produces a 202 Accepted response. Optionally, also filters on request content type.",
     icon: 'sign-in',
-    resolvePort: {
-        name: 'out',
-        description: "Request body as a String",
-        datatype: 'string'
-    },
-    rejectPort: {
-        name: 'error',
-        description: "Error object",
-        datatype: 'object'
-    },
     outPorts: {
         rejected: {
             description: "HTTP 415 Unsupported Media Type request/response pair {req, res}",
--- a/components/http-accept.js
+++ b/components/http-accept.js
@@ -50,7 +50,7 @@ function handle(pair){
     if (contentTypeMatches(this.types, pair.req.headers['content-type'])) {
         if (_.has(pair.req, 'body')) {
             if (outPorts.accepted.isAttached()) {
-                pair.res.writeHead(202);
+                pair.res.writeHead(202, "Accepted");
                 pair.res.write("Accepted");
                 outPorts.accepted.send(pair);
                 outPorts.accepted.disconnect();
@@ -63,7 +63,7 @@ function handle(pair){
                 encoding: charset || this.encoding || 'utf8'
             }).then(function(body){
                 if (outPorts.accepted.isAttached()) {
-                    pair.res.writeHead(202);
+                    pair.res.writeHead(202, "Accepted");
                     pair.res.write("Accepted\n");
                     outPorts.accepted.send(pair);
                     outPorts.accepted.disconnect();
@@ -71,7 +71,7 @@ function handle(pair){
                 return body;
             }, function(err){
                 if (outPorts.rejected.isAttached()) {
-                    pair.res.writeHead(413);
+                    pair.res.writeHead(413, "Payload Too Large");
                     pair.res.write(err.message);
                     outPorts.rejected.send(pair);
                     outPorts.rejected.disconnect();
@@ -81,7 +81,7 @@ function handle(pair){
         }
     } else {
         if (outPorts.rejected.isAttached()) {
-            pair.res.writeHead(415);
+            pair.res.writeHead(415, "Unsupported Media Type");
             pair.res.write("Expected " + this.types + " not " + pair.req.headers['content-type']);
             outPorts.rejected.send(pair);
             outPorts.rejected.disconnect();
--- a/components/http-accept.js
+++ b/components/http-accept.js
@@ -1,55 +1,59 @@
 // http-accept.js
 
 var _ = require('underscore');
-var body = require('body');
-var noflo = require('noflo');
+var getRawBody = require('raw-body');
+var typer = require('media-typer');
 
 var basenode = require('./base-node');
+var promiseComponent = require('./promise-component');
 
-exports.getComponent = function() {
-    return _.extend(new noflo.Component({
-        outPorts: {
-            out: {
-                description: "Request body as a String",
-                datatype: 'string'
-            },
-            rejected: {
-                description: "HTTP 415 Unsupported Media Type request/response pair {req, res}",
-                datatype: 'object'
-            },
-            accepted: {
-                description: "HTTP 202 Accepted request/response pair {req, res}",
-                datatype: 'object'
-            }
+exports.getComponent = promiseComponent({
+    description: "Extracts the request body as a String and produces a 202 Accepted response. Optionally, also filters on request content type.",
+    icon: 'sign-in',
+    resolvePort: {
+        name: 'out',
+        description: "Request body as a String",
+        datatype: 'string'
+    },
+    rejectPort: {
+        name: 'error',
+        description: "Error object",
+        datatype: 'object'
+    },
+    outPorts: {
+        rejected: {
+            description: "HTTP 415 Unsupported Media Type request/response pair {req, res}",
+            datatype: 'object'
         },
-        inPorts: {
-            limit: {
-                description: "Request Content-Length limit",
-                datatype: 'int',
-                process: basenode.on({data: basenode.assign('limit')})
-            },
-            encoding: {
-                description: "Request body character encoding",
-                datatype: 'string',
-                process: basenode.on({data: basenode.assign('encoding')})
-            },
-            type: {
-                description: "Request Content-Type",
-                datatype: 'string',
-                process: basenode.on({data: basenode.assign('types', basenode.push)})
-            },
-            'in': {
-                description: "HTTP request/response pair {req, res}",
-                datatype: 'object',
-                required: true,
-                process: basenode.on({data: handle})
-            }
+        accepted: {
+            description: "HTTP 202 Accepted request/response pair {req, res}",
+            datatype: 'object'
         }
-    }), {
-        description: "Extracts the request body as a String and produces a 202 Accepted response. Optionally, also filters on request content type.",
-        icon: 'sign-in'
-    });
-};
+    },
+    inPorts: {
+        limit: {
+            description: "Request Content-Length limit",
+            datatype: 'int',
+            ondata: basenode.assign('limit')
+        },
+        encoding: {
+            description: "Request body character encoding",
+            datatype: 'string',
+            ondata: basenode.assign('encoding')
+        },
+        type: {
+            description: "Request Content-Type",
+            datatype: 'string',
+            ondata: basenode.assign('types', basenode.push)
+        },
+        'in': {
+            description: "HTTP request/response pair {req, res}",
+            datatype: 'object',
+            required: true,
+            ondata: handle
+        }
+    }
+});
 
 function handle(pair){
     var outPorts = this.outPorts;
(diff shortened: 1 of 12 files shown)


components/http-accept.js
```diff
@@ -1,55 +1,59 @@
...
-                datatype: 'object'
-            },
-            accepted: {
-                description: "HTTP 202 Accepted request/response pair {req, res}",
-                datatype: 'object'
-            }
+exports.getComponent = promiseComponent({
+    description: "Extracts the request body as a String and produces a 202 Accepted response. Optionally, also filters on request content type.",
+    icon: 'sign-in',
+    resolvePort: {
```
glenna-m: Why do we need to specify the resolvePort and rejectPort here?  You do not use them in this component as far as I can tell.  Should they be part of the promiseComponent instead?

components/http-accept.js
```diff
@@ -1,55 +1,59 @@
...
-                description: "HTTP 415 Unsupported Media Type request/response pair {req, res}",
-                datatype: 'object'
-            },
-            accepted: {
-                description: "HTTP 202 Accepted request/response pair {req, res}",
-                datatype: 'object'
-            }
+exports.getComponent = promiseComponent({
+    description: "Extracts the request body as a String and produces a 202 Accepted response. Optionally, also filters on request content type.",
+    icon: 'sign-in',
```
glenna-m, line 12: Curious why you chose the sign-in icon?   Is there no page type icon?

components/http-accept.js
```diff
@@ -61,30 +65,28 @@ function handle(pair){
...
+                if (outPorts.accepted.isAttached()) {
+                    pair.res.writeHead(202);
+                    pair.res.write("Accepted\n");
+                    outPorts.accepted.send(pair);
+                    outPorts.accepted.disconnect();
+                }
+                return body;
+            }, function(err){
+                if (outPorts.rejected.isAttached()) {
+                    pair.res.writeHead(413);
```
glenna-m: I think this code would be a little cleaner if we pick up the http-status-codes package and use their status code constants.  It's a little more readable and standard.

components/promise-component.js
```diff
@@ -0,0 +1,66 @@
+// promise-component.js
+
+var _ = require('underscore');
+var noflo = require('noflo');
+
+module.exports = function(def){
+    if (_.isEmpty(def.inPorts)) throw Error("Missing inPorts");
+    if (def.resolvePort && !_.isString(def.resolvePort.name))
```
glenna-m, line 9: Are you sure you want these (resolve/reject ports) defined outside the promiseComponent?   If they were part of this component, you could skip the error checking here.

components/promise-component.js
```diff
@@ -0,0 +1,66 @@
...
+        throw Error("Missing resolvePort name");
+    if (def.rejectPort && !_.isString(def.rejectPort.name))
+        throw Error("Missing rejectPort name");
+    var resolvePort = def.resolvePort || _.defaults({
```
glenna-m, line 15: This seems odd - why do you check to see if the reject/resolve port is defined and has a name above (and throw error if not), and then set these defaults on it below with a hard coded default name?

components/promise-component.js
```diff
@@ -0,0 +1,66 @@
...
+    }, _.propertyOf(def.outPorts)('error')) || {
+        name: 'error',
+        description: "Rejected error of the updater",
+        datatype: 'object',
+        addressable: false,
+        buffered: false,
+        required: false,
+        caching: false
+    };
+    var inPorts = _.mapObject(def.inPorts, function(port, name){
```
glenna-m, line 38: Should this function have a socketIndex parameter on the end of it?

components/promise-component.js
```diff
@@ -0,0 +1,66 @@
...
+        var process = port.process || function(event, payload) {
+            var key = 'on' + event;
+            if (port[key]) return port[key].call(this.nodeInstance, payload);
+        };
+        return _.defaults({
+            process: function(event, payload, socketIndex) {
+                var self = this;
+                var outPorts = this.nodeInstance.outPorts;
+                Promise.resolve().then(function(){
+                    return process.call(self, event, payload, socketIndex);
```
glenna-m, line 48: Calling with process with socketIndex, but I don't see that parameter on the default definition for it above.  Am I missing something?

components/rdf-load.js
```diff
@@ -3,53 +3,48 @@
...
-                description: "All the quads as a RDF JS Interface Graph object",
-                datatype: 'object'
-            },
-            error: {
-                description: "Error object",
-                datatype: 'object'
-            }
+exports.getComponent = promiseComponent({
+    description: "Loads data into a RDF JS Interface Graph object",
+    icon: 'sign-in',
```
glenna-m, line 12: sign-in icon seems like a strange one for an rdf-load component.

components/http-accept.js
catch-point: I added them to include a description, but I take your point, and I'll move the result description to the main component description.

components/http-accept.js
catch-point, line 12: I thought the arrow going into a box was an appropriate imagery for this (nothing to do with the action of signing-in though).

components/promise-component.js
catch-point, line 9: This gives the caller the ability to use custom names for resolve/reject ports.

components/promise-component.js
catch-point, line 15: I'm not sure I understand your question. The code checks that if the caller provided a custom resolve/reject port, they must also provide a name; otherwise this code provides default resolve/reject ports.

components/promise-component.js
catch-point, line 38: No this function is just looping through all the inPorts (not handling data).

components/promise-component.js
catch-point, line 48: The default process does not use socketIndex, but port.process (not defined here) might.

components/http-accept.js
catch-point: I'm not sure how that would make it cleaner. Can you show me what you mean?

graphs/0035_GraphNode_Basic.json
```diff
@@ -30,7 +30,7 @@
       "component": "rdf-components/RequestTemplate",
       "metadata": {
         "label": "LOAD INTO GRAPH",
-        "x": 324,
+        "x": 288,
```
glenna-m, line 33: Would like to see us move this graph 35 json file over to the noflo-rdf-pipeline/tests since it really should belong with David's legacy tests there.

components/rdf-load.js
catch-point, line 12: The imagery of an arrow pointing into a box seemed appropriate for this parsing functionality. I thought the upload/download icons carried too much semantic weight.

graphs/0035_GraphNode_Basic.json
catch-point, line 33: I think there should be something here to document/demonstrate how to use a request template, but perhaps a more generic example would be better.

test/promise-component-mocha.js
```diff
@@ -0,0 +1,149 @@
...
+                        handler(payload);
+                    }
+                }
+            }
+        }).then(promiseComponent).then(createComponent).then(function(component){
+            return new Promise(function(callback){
+                handler = callback;
+                sendData(component, 'in', "hello");
+            });
+        }).should.become("hello");
```
glenna-m, line 45: Not entirely sure how the promise to create a component and then send "hello" data to the input port's hander becomes "hello".  Can you add a comment to indicate what this should.become.hello check is looking at to find the "hello" string to make this a little easier to follow?

glenna-m:
Really nice work.

components/http-accept.js
glenna-m, line 12: OK.  Make sense.

components/http-accept.js
glenna-m: You are using hard coded values like 202 or 413 for your header status.  Would rather see something like HttpStatus.ACCEPTED or HttpStatus.REQUEST_TOO_LONG since it's more clear without having to look them up.

components/promise-component.js
glenna-m, line 9: OK.  Maybe add a comment to this effect?

components/promise-component.js
glenna-m, line 15: As noted above, a comment as to what you were thinking here would be really helpful.   No issue then with the code.

components/promise-component.js
glenna-m, line 48: Why are you leaving the socket index off the default process?

components/promise-component.js
glenna-m, line 38: Right.  Was on the wrong line with my comment.  The following comment is what I meant.  No issue here.

components/rdf-load.js
glenna-m, line 12: OK

test/promise-component-mocha.js
catch-point, line 45: DONE

graphs/0035_GraphNode_Basic.json
glenna-m, line 33: Yes.  Though I wouldn't lose this graph.  It fits well with the rdf-pipeline ported tests.   Would like to see it put with them so someone looking at the others sees this one too.

test/promise-component-mocha.js
glenna-m, line 45: Let's talk about this.  I think I am still missing something (I followed what the comment says before), but it's probably more a question for me than the code here.

components/http-accept.js
```diff
@@ -1,55 +1,49 @@
...
+            description: "Request Content-Length limit",
+            datatype: 'int',
+            ondata: basenode.assign('limit')
+        },
+        encoding: {
+            description: "Request body character encoding",
+            datatype: 'string',
+            ondata: basenode.assign('encoding')
+        },
+        type: {
```
dbooth-boston, line 34: How about naming the ports based on the way that the corresponding HTTP headers are named (except capitalization), so that the correspondence is clear?

components/http-accept.js
```diff
@@ -1,55 +1,49 @@
...
+            description: "Request Content-Type",
+            datatype: 'string',
+            ondata: basenode.assign('types', basenode.push)
```
dbooth-boston, line 37: When we get to it, I think the "ondata" property should be removed from all of the port definitions here, because (except for special components) all ports should automatically use the same function for handling data events.  However, we don't have to make that change in this PR.  It can be later.

components/http-accept.js
catch-point, line 34: What headers would those be?

components/promise-component.js
```diff
@@ -0,0 +1,66 @@
+// promise-component.js
+
+var _ = require('underscore');
+var noflo = require('noflo');
+
+module.exports = function(def){
```
dbooth-boston: Could one of you please explain module.exports to me sometime, in the case when a function is assigned?  Web searches didn't add much light.

components/rdf-jsonld.js
```diff
@@ -3,53 +3,33 @@
...
-            out: {
-                description: "JSON LD Graph object",
-                datatype: 'object'
-            },
-            error: {
-                description: "Error object",
-                datatype: 'object'
-            }
+exports.getComponent = promiseComponent({
+    description: "Converts an RDF JS Interface Graph object into a JSON LD Graph object",
```
dbooth-boston, line 11: What is an RDF JS Interface graph?

components/rdf-update.js
```diff
@@ -4,65 +4,51 @@ var _ = require('underscore');
...
-                process: basenode.on({data: execute})
-            }
+        update: {
+            description: "SPARQL update template string in handlebars syntax",
+            datatype: 'string',
+            required: true,
+            ondata: basenode.assign('update', Handlebars.compile)
+        },
+        "default": {
+            description: "Graph URI for the default dataset",
```
dbooth-boston: What do you mean by "default dataset"?  Do you mean repository?

test/promise-component-mocha.js
```diff
@@ -0,0 +1,151 @@
...
+        _.forEach(component.outPorts, function(port, name) {
+            port.nodeInstance = component;
+            port.name = name;
+        });
+        return component;
+    }
+    function sendData(component, port, payload) {
+        var socket = noflo.internalSocket.createSocket();
+        component.inPorts[port].attach(socket);
+        socket.send("hello");
```
glenna-m: Why is this sending a hard coded "hello" instead of the payload?

components/promise-component.js
```diff
@@ -0,0 +1,66 @@
...
+    if (_.isEmpty(def.inPorts)) throw Error("Missing inPorts");
+    if (def.resolvePort && !_.isString(def.resolvePort.name))
+        throw Error("Missing resolvePort name");
+    if (def.rejectPort && !_.isString(def.rejectPort.name))
+        throw Error("Missing rejectPort name");
+    var resolvePort = def.resolvePort || _.defaults({
+        name: 'out'
+    }, _.propertyOf(def.outPorts)('out')) || {
```
dbooth-boston, line 17: After states have been implemented, there will be a state for 'out', right?

components/promise-component.js
dbooth-boston, line 48: Except perhaps for special components, I think we should be moving toward all ports using the same 'process' function.   Different logic needed by different wrappers (component types) should be moved into the wrapper functions.  See https://github.com/rdf-pipeline/framework/wiki/Wrapper-Interface  The logic specific to each wrapper (component type) should be in that wrapper's fRunUpdater function.  We should discuss this.

components/rdf-jsonld.js
catch-point, line 11: https://www.w3.org/TR/2011/WD-rdf-interfaces-20110510/#graphs

components/rdf-update.js
catch-point: The default Graph (like in a FROM clause) https://www.w3.org/TR/sparql11-query/#unnamedGraph

test/promise-component-mocha.js
catch-point: FIXED

components/rdf-update.js
catch-point: Changed to use the term Graph instead.

components/promise-component.js
catch-point, line 17: yes, but that's not for this PR

components/http-accept.js
dbooth-boston, line 34: I was thinking Content-Length (-> content_length_limit), Accept-Encoding, Content-Type.  But maybe we should try it out more first, both because that renaming would make the port names longer and because you have the port descriptions readily available.   We can leave it as is for the moment anyway.

PR #6, opened 2016-01-21
New JSON Splitter component:

glenna-m:
- splits JSON input into one or more records
- generates a vnid for each record using the configured json pointer
- outputs each record for consumption by the next component

### DIFF
--- a/components/base-node.js
+++ b/components/base-node.js
@@ -7,6 +7,8 @@
 
 var _ = require('underscore');
 
+var promiseComponent = require('./promise-component');
+
 module.exports = {
 
   /**
@@ -65,7 +67,48 @@ module.exports = {
        ar.push(item);
      }
      return ar;
-   } 
+   },
+
+   /**
+    * Gets a promise to build a no-flo component around a user-defined rdf pipline updater.
+    *
+    * @param def updater definition
+    */
+   updaterComponent: function(def) {
+
+     // Get a promise to return a noflo component 
+     return promiseComponent( 
+
+        // Set the defaults for the component
+        _.defaults(
+          def,
+          { 
+            inPorts: 
+              // add ondata handler function to every port
+              _.mapObject(
+                def.inPorts,
+                function(port, portName) {
+
+                  return _.defaults(
+                           port,
+                           {
+                             // define the ondata handler
+                             ondata: function ( payload, socketIndex ) {
+                               this[portName] = payload;
+                               var self = this;
+                               var keys = _.keys( def.inPorts );
+                               var shouldUpdate = 
+                                 _.reduce( keys, 
+                                           function( memo, key ) { return ( memo && ! _.isUndefined( self[key] )); }, 
+                                           true );
+                               if ( shouldUpdate ) {
+                                 return def.update.call( this, _.pick( this, _.keys(def.inPorts)));
+                               }
+                             }
+                         }) // _.defaults on each port
+              }) // mapObject
+        }) // defaults
+     ); // promise
+   }
 
 };
-
--- a/components/json-splitter.js
+++ b/components/json-splitter.js
@@ -1,0 +1,85 @@
+// json-splitter.js
+
+/**
+ * This component takes a json string with one or more array elements, parses it, and makes each array element into an 
+ * individual records.  Each record is adorned with a virtual node ID (vnid) to uniquely identify it in the rdf pipeline, 
+ * and sends them on to the next component.
+ */
+
+var jsonpointer = require('jsonpointer');
+
+var _ = require('underscore');
+var basenode = require('./base-node');
+
+exports.getComponent = basenode.updaterComponent({
+
+  description: 'Splits a json dataset into individual records, adds a vnid to each, and sends them on.',
+
+  inPorts: { 
+    in: { 
+      addressable: true, 
+      datatype: 'string',
+      description: 'json string input',
+      required: true
+    },
+
+    json_pointer: {
+      addressable: true, 
+      datatype: 'string',
+      description: 'json pointer for the json data attribute to use as the vnid (RFC 6901)',
+      required: true
+    }
+  },
+
+  update: function( args ) {   
+
+    var outPort = this.outPorts.out; 
+    var errorPort = this.outPorts.error; 
+
+    // Attempt to parse the JSON data received
+    var parsedData = null;
+    try { 
+      parsedData = JSON.parse( this.in.replace(/\r?\n|\r/,'') );
+    } catch(e) {
+      basenode.handleError( "Unable to parse JSON for "+this.in+": "+e ); 
+    }
+
+    var key = null;
+    var didSend = false;
+    if ( _.isArray( parsedData ) ) { 
+
+      var jsonPointer = jsonpointer.compile(this.json_pointer); 
+      parsedData.forEach( function(datum) {   
+
+        // Get and assign the hash key for this record
+        key = jsonPointer.get(datum);
+        if ( _.isNull( key ) ) {
+          this.handleError( "Unable to find "+args.json_pointer+" in "+datum);
+        } else {
+          datum['vnid'] = key;
+          outPort.send(datum); 
+          didSend = true;
+        }
+
+      }); 
+      if ( didSend ) { outPort.disconnect(); }
+
+    } else if ( _.isObject( parsedData ) ) { 
+
+       // Have just one element in the json data, so build our one element hash
+       key = jsonpointer.get(parsedData, args.json_pointer);
+       if ( _.isNull( key ) ) {
+         this.handleError( 'Unable to find '+this.json_pointer+' in '+parsedData );
+       } else {
+         parsedData['vnid'] = key;
+         outPort.send(parsedData); 
+         outPort.disconnect(); 
+       }
+
+    } else {
+      this.handleError('Expected a JSON object or array of objects!');
+    }
+
+  }
+
+});
--- a/test/json-splitter-mocha.js
+++ b/test/json-splitter-mocha.js
@@ -1,0 +1,228 @@
+/**
+ * File: jsonSplitter.js
+ * Unit tests for the APIs defined in components/jsonSplitter.js
+ */
+
+var chai = require('chai');
+
+var chaiAsPromised = require('chai-as-promised');
+chai.use(chaiAsPromised);
+
+var expect = chai.expect;
+var should = chai.should();
+
+var sinon = require('sinon');
+
+var _ = require('underscore');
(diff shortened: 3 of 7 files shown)


components/base-node.js
```diff
@@ -44,6 +46,18 @@ module.exports = {
    },
 
   /**
+   * Standard component error handling - send a message to the error port and throw Error
+   *
+   * @param message message to be sent to error port
+   * @param errorPort handle to the component's error port
+   */
+   handleError: function( message, errorPort ) {
```
dbooth-boston: I think we're going to want to move to a model of logging errors, but I'm okay with keeping this as is for the moment, at least until states are implemented.

glenna-m: It is up to the user (e.g, Alice), whether they choose to hook up the error port.  They don't have to.  If they do not, the clutter is minimal.    I have started to centralize my error handling in that new baseNode.handleError API - should make it easy to add logging there.

components/base-node.js
```diff
@@ -65,7 +79,32 @@ module.exports = {
        ar.push(item);
      }
      return ar;
-   } 
+   },
```
dbooth-boston, line 70: I'm curious why the 'push' function is being defined above on line 62/76.  To automatically create an array if there isn't one already?

glenna-m, line 70: That is a simple helper function that James was using in multiple places to dynamically build up his incoming port data.    Rather than duplicate it in multiple components, I centralized it here. Look at rdf-query and rdf-update for good examples of the use case.

components/base-node.js
```diff
@@ -44,6 +46,18 @@ module.exports = {
    },
 
   /**
+   * Standard component error handling - send a message to the error port and throw Error
```
dbooth-boston: FYI, the 'assign' function above (line 17/19) will go away when we get a uniform wrapper API implemented, as I mentioned in email.

components/base-node.js
```diff
@@ -65,7 +79,32 @@ module.exports = {
        ar.push(item);
      }
      return ar;
-   } 
+   },
+
+   updatePromise: function(def) {
```
dbooth-boston: A comment before each function would be helpful, describing its purpose.

components/base-node.js
```diff
@@ -65,7 +79,32 @@ module.exports = {
...
+             def,
+             { inPorts: _.mapObject(
+                          def.inPorts,
+                          function(port, portName) {
+
+                            return _.defaults(
+                                     port,
+                                     {
+                                       ondata: function ( payload, socketIndex ) {
+                                                 this[portName] = payload;
```
dbooth-boston: What kind of object is 'this' supposed to be bound to, at this point?  I'm trying to figure out why ondata is assigning the payload to this[portName].   Is it because states are not yet implemented?

glenna-m: This will be changing.  I actually removed some of my latest vni code from this API in an attempt to keep this review focused on the json_splitter component without getting into the vni manager that will be called from here.

The updatePromise API is used to extend the generic promiseComponent to call the component's onchange API.  Without it, you wouldn't get the custom component data handling. 

At the time ondata function is executed, we will be running on the node context.

components/json-splitter.js
```diff
@@ -0,0 +1,88 @@
...
+
+          var self = this;
+
+          var outPort = this.outPorts.out; 
+          var errorPort = this.outPorts.error; 
+
+          // Attempt to parse the JSON data received
+          var parsedData = null;
+          try { 
+             parsedData = JSON.parse( this.in.replace(/\r?\n|\r/,'') );
```
dbooth-boston: Is this being done because state management is not yet implemented?

components/json-splitter.js
```diff
@@ -0,0 +1,88 @@
...
+        addressable: true, 
+        datatype: 'string',
+        description: 'json pointer for the json data attribute to use as the vnid (RFC 6901)',
+        required: true
+      }
+  },
+
+  onchange: function( data ) {   
+
+      if ( this.in && this.json_pointer ) {
```
dbooth-boston: What is 'this' expected to be at this point?  An inPorts object?  But below it seems like a NoFlo component instance.  I don't understand how there is both this.in and this.outPorts .  I would think it would be this.inPorts.in .

glenna-m: They are two different things: 
- this.in is an attribute that contains whatever content was sent to the "in" port, e.g., a json string 
- this.inPorts is the noflo inPorts object itself.   This is set when we construct the component.

To explain the this.in (and this.json_pointer), the promise-component sets up a callback for when we get incoming data (the process call) on a port.  When we get incoming port data, process will lookup the ondata function for the port and call it.  In the case of this component, it will be calling the ondata function that was defined in updatePromise.  The very first line of that ondata function is setting the this.in (and this.json_pointer when called on that port):   this[portName] = payload;

This.inPorts.in and this.inPorts.json_pointer are also around, but they don't have the incoming port data on them. They have the noflo port definitions.

Does this help?

components/json-splitter.js
```diff
@@ -0,0 +1,88 @@
...
+
+  inPorts: { 
+      in: { 
+        addressable: true, 
+        datatype: 'string',
+        description: 'json string input',
+        required: true
+      },
+
+      json_pointer: {
```
dbooth-boston: Do we need json_pointer?  The split component should receive the vnids as a hash or list directly.  I'd like to keep this component as simple as possible.

components/base-node.js
catch-point: We may want to (outside this PR) to produce low level components that reflect their technical nature (i.e. they only do one thing) and produce corresponding higher level subgraphs that capture common use cases for them (such as having a logger already attached into their error port or an http-accept attached to an inport).

components/base-node.js
catch-point: Can you add a validation to ensure def.onchange is a function?

components/json-splitter.js
catch-point: It will receive the data on the 'in' port, the json_pointer is the attribute name (or path) that contains the vnid, such as 'key' or 'id'

components/base-node.js
catch-point: This function returns a component factory function. Can we improve the name to better indicate this, such as updaterComponent?

components/base-node.js
```diff
@@ -65,7 +79,32 @@ module.exports = {
...
+                                                 var self = this;
+                                                 return def.onchange.call( this, _.pick( this, _.keys(def.inPorts)));
```
catch-point: This onchange function should only be fired if all the required inPorts have a received a  state.

components/json-splitter.js
catch-point: This check (if both states have been received) should be done before the onchange event is fired (i.e. in base-node).

components/json-splitter.js
```diff
@@ -0,0 +1,88 @@
...
+
+          var outPort = this.outPorts.out; 
+          var errorPort = this.outPorts.error; 
+
+          // Attempt to parse the JSON data received
+          var parsedData = null;
+          try { 
+             parsedData = JSON.parse( this.in.replace(/\r?\n|\r/,'') );
+          } catch(e) {
+            baseNode.handleError( "Unable to parse JSON for "+this.in+": "+e,
```
catch-point: This is causing the error to be reported twice. Once by the handleError and again by the promise-component.

s/baseNode.handleError/throw Error/

components/base-node.js
catch-point: I don't see the point of this function, whenever I test it it causes the error to be reported twice.

components/json-splitter.js
```diff
@@ -0,0 +1,88 @@
...
+
+              var jsonPointer = jsonpointer.compile(this.json_pointer); 
+              parsedData.forEach( function(datum) {   
+
+                  // Get and assign the vnid for this record
+                  vnid = jsonPointer.get(datum);
+                  if ( _.isNull( vnid ) ) 
+                      throw new Error( "Unable to find "+self.json_pointer+" in "+datum,
+                                       errorPort); 
+                  datum['vnid'] = vnid;
```
catch-point: I think the metadata (such as vnid) should be separated from the state (datum). Perhaps this should be changes to 
  outPort.send({vnid: vnid, state: datum})

test/json-splitter-mocha.js
```diff
@@ -0,0 +1,225 @@
...
+  });
+
+  describe('#onchange', function() {
+
+    it('should have onchange API', function() {
+        jsonSplitter.getComponent.should.exist;
+        jsonSplitter.getComponent.should.be.a('function');
+    });
+
+    it('should generate a record for each element of json input data array with a vnid', function( done ) {
```
jamesrdf: The done callback is not needed if the function returns a Promise.

components/base-node.js
catch-point: What do you think going through the component's EventEmitter API to trigger onchange?

components/json-splitter.js
glenna-m: Yes.  I did that in the json_hasher.  Will bring that into this one if we keep it.

components/base-node.js
glenna-m: The point is to centralize error handling so it is easy to modify the algorithm used.  If you are seeing errors reported twice, do you have the configuration hooked to an output component?    David wants to get away from using the noflo component output and do logging without it to simplify the edges.

glenna-m: BTW, the location of handleErrors shifted in PR 7.  I have added it to the promiseComponent extension to make it easy for all components to use it.

components/base-node.js
glenna-m: I have added the check for all ports having received a state before calling the updater.  Not sure about the EventEmitter API - let's talk about it.

components/json-splitter.js
glenna-m: HandleError does more than throw the error.  It is also sending the data to the error port and should be logging.  Let's talk how we want error handling to work.

components/json-splitter.js
glenna-m: Makes sense.  Will do this.

components/base-node.js
```diff
@@ -65,7 +67,48 @@ module.exports = {
...
+
+                  return _.defaults(
+                           port,
+                           {
+                             // define the ondata handler
+                             ondata: function ( payload, socketIndex ) {
+                               this[portName] = payload;
+                               var self = this;
+                               var keys = _.keys( def.inPorts );
+                               var shouldUpdate =
```
catch-point, line 100: shouldUpdate should ignore non-required ports

PR #7, opened 2016-01-22
json-hasher with demo'ing updater interface ideas

glenna-m:


components/base-node.js
```diff
@@ -65,7 +67,43 @@ module.exports = {
...
+                          function(port, portName) {
+
+                            return _.defaults(
+                                     port,
+                                     {
+                                       ondata: function ( payload, socketIndex ) {
+                                                 this[portName] = payload;
+                                                 var self = this;
+                                                 var shouldUpdate = 
+                                                      _.reduce( _.keys( def.inPorts ),
```
catch-point, line 90: This should only apply to required ports

components/promise-component.js
```diff
@@ -60,11 +60,23 @@ module.exports = function(def){
...
+
             outPorts: _.defaults(_.object(
                 [resolvePort.name, rejectPort.name],
                 [_.clone(resolvePort), _.clone(rejectPort)]
             ), _.mapObject(def.outPorts, _.clone)),
             inPorts: _.mapObject(inPorts, _.clone)
-        }), _.pick(def, 'description', 'icon'));
+        }), 
+        _.pick(def, 'description', 'icon'), 
+        { handleError:  function(message) {
```
catch-point, line 71: This seems unnecessary, it looks like it would produce the error three times!

package.json
```diff
@@ -47,8 +49,9 @@
       "http-accept": "./components/http-accept.js",
       "round-robin": "./components/round-robin.js",
       "file-node": "./components/file-node.js",
+      "js-node": "./components/js-node.js",
       "json-file-node": "./components/json-file-node.js",
-      "js-node": "./components/js-node.js"
+      "json-hasher": "./components/json-hasher.js"
```
catch-point, line 54: Where is this file?

components/json-hasher.js
```diff
@@ -0,0 +1,89 @@
...
+   *   - file path to json data to process
+   *   - json pointer describing the key to extract from each json data record and 
+   *     use as the hash key for the record.
+   */
+  update: function( args ) {   
+
+      // Attempt to read and parse the JSON file 
+      var parsedData = null;
+      try { 
+        var jsonString = fs.readFileSync( args.file_path ).toString();
```
catch-point, line 48: I think it would be better to have a separate component that reads json from a file (not here in this component)

components/base-node.js
```diff
@@ -65,7 +67,43 @@ module.exports = {
        ar.push(item);
      }
      return ar;
-   } 
+   },
+
+   updatePromise: function(def) {
```
catch-point, line 72: I think we should avoid using the term update or updater at this layer to avoid confusion with the wrappers. Instead maybe we should call this something like vniComponent to indicate that it will be responsible for managing the vni states.

components/json-hasher.js
glenna-m, line 48: Spoke with David about this.  He sees this component as mostly useful for demo purposes.  Handling a file here means we only need to configure and show one component instead of 3  (read file, parser, hasher) -> the recommendation is to keep the file handling in here.

components/base-node.js
glenna-m, line 72: This will change in the next check in.

PR #8, opened 2016-01-22
If no socket is attached to error port, log to console.error

catch-point:


test/promise-component-mocha.js
```diff
@@ -118,6 +118,32 @@ describe('promise-component', function() {
             return onceData(component, 'outgoing', 'sentback', 'incoming', "hello");
         }).should.be.rejectedWith("hello world");
     });
+    it("should log rejections to console.error when no reject port attached", function() {
```
glenna-m, line 121: I do not follow how you are forcing the reject port to not be attached.  That port is created inside promise-component, no?

test/promise-component-mocha.js
```diff
@@ -118,6 +118,32 @@ describe('promise-component', function() {
...
+                'in':{
+                    ondata: function(payload) {
+                        throw payload + " world";
+                    }
+                }
+            }
+        }).then(promiseComponent).then(createComponent).then(function(component){
+            var console_error = console.error;
+            return new Promise(function(resolve, reject) {
+                console.error = reject;
```
glenna-m, line 133: Also do not follow how this really tests console logging.  Perhaps we should go over this on Monday.

test/promise-component-mocha.js
catch-point, line 121: The outport is created, but in this test nothing is attached to it (only the in/out ports are attached).

glenna-m:
After discussion with James, this looks good.  LGTM to merge.

PR #9, opened 2016-01-25
Add state-component

catch-point:
This code manages state of inPorts

### DIFF
--- a/components/promise-component.js
+++ b/components/promise-component.js
@@ -38,7 +38,7 @@ module.exports = function(def){
     var inPorts = _.mapObject(def.inPorts, function(port, name){
         var process = port.process || function(event, payload, socketIndex) {
             var key = 'on' + event;
-            if (port[key]) return port[key].call(this.nodeInstance, payload);
+            if (port[key]) return port[key].call(this.nodeInstance, payload, socketIndex);
         };
         return _.defaults({
             process: function(event, payload, socketIndex) {
--- a/components/state-component.js
+++ b/components/state-component.js
@@ -56,7 +56,7 @@ function fireChangeEvent(required, onchange) {
     return function(key) {
         var wild = _.pick(this.inPayloads, isWild.bind(this));
         if (_.isUndefined(key)) {
-            // no key for this event, first with only wild matching payloas
+            // no key for this event, fire with only wild matching payloads
             var missing = _.difference(required, _.keys(wild));
             if (_.isEmpty(missing)) {
                 var self = this;
--- a/components/state-component.js
+++ b/components/state-component.js
@@ -5,16 +5,18 @@ var jsonpointer = require('jsonpointer');
 var promiseComponent = require('./promise-component');
 
 /**
- * Stores the last packet payload for every inPort and calls an onchange function
- * when received and after when all required port have received something.
- * If multiple ports both provide an indexBy property name, path or function,
- * only packets that both resolve to the same key with be called together.
+ * This combines payloads, from multiple ports, arriving in an asynchronous
+ * fashion. Each port may include an indexBy function, property, or path
+ * that will be used to determine a key that must match payloads from other ports.
+ * If no indexBy is provided then the payload will match any payload from other ports.
+ * A provided onchange function is called when payload is received (from all
+ * required port) with payload having the same key from all other inPorts.
  *
  * Usage:
  *   stateComponent({
- *     inPorts: {name:{indexBy:function(paylod)}} of port names to port options || [] of port names,
+ *     inPorts: {name:{indexBy:function(payload)}} of port names to port options || [] of port names,
  *     onchange: function(inPayloads, outPayload) {
- *         inPayloads = {} of port names to last payload,
+ *         inPayloads = {name:payload} of port names to last payload,
  *         outPayload = previous result of onchange
  *     }
  *   });
@@ -27,28 +29,33 @@ module.exports = function(def){
     return promiseComponent(_.defaults({
         inPorts: _.mapObject(inPorts, function(port, name) {
             return _.extend({
-                ondata: _.compose(change(required, def.onchange), data(port && port.indexBy, name))
+                ondata: _.compose(
+                    fireChangeEvent(required, def.onchange),
+                    indexPayload(port && port.indexBy, name)
+                )
             }, port);
         }),
     }, def));
 };
 
-function data(indexBy, name) {
+function indexPayload(indexBy, name) {
     var by = _.isFunction(indexBy) ? indexBy : pointer(indexBy);
     return function(payload, socketIndex) {
+        // determine a key string or undefined if no key should be used
         var key = indexBy == null ? undefined : ('' + by.call(this, payload));
         var payloads = this.inPayloads || {};
-        var index = indexBy == null ? payload :
+        var index = _.isUndefined(key) ? payload :
             _.extend(payloads[name] || {}, _.object([key], [payload]));
         this.inPayloads = _.extend(payloads, _.object([name], [index]));
         return key;
     };
 }
 
-function change(required, onchange) {
+function fireChangeEvent(required, onchange) {
     return function(key) {
         var wild = _.pick(this.inPayloads, isWild.bind(this));
         if (_.isUndefined(key)) {
+            // no key for this event, first with only wild matching payloas
             var missing = _.difference(required, _.keys(wild));
             if (_.isEmpty(missing) && _.isFunction(onchange)) {
                 var self = this;
@@ -57,6 +64,7 @@ function change(required, onchange) {
                 });
             }
         } else {
+            // a particular key must match payload (or be wild matching)
             var payloads = _.extend(_.mapObject(this.inPayloads, _.property(key)), wild);
             var missing = _.difference(required, _.keys(payloads));
             if (_.isEmpty(missing) && _.isFunction(onchange)) {
--- a/components/state-component.js
+++ b/components/state-component.js
@@ -1,0 +1,49 @@
+// state-component.js
+
+var _ = require('underscore');
+var promiseComponent = require('./promise-component');
+
+/**
+ * Stores the most recent packet for every inPort and calls an onchange function
+ * when they change and after when all required port have received something.
+ * If multiple ports both provide an indexBy property name or function,
+ * only packets that both resolve to the same value with be called together
+ */
+module.exports = function(def){
+    var inPorts = _.isArray(def.inPorts) ? _.object(def.inPorts, []) : def.inPorts;
+    var required = _.compact(_.map(inPorts, function(port, name){
+        return port && port.required && name;
+    }));
+    return promiseComponent(_.defaults({
+        inPorts: _.mapObject(inPorts, function(port, name) {
+            return _.extend({
+                ondata: _.compose(change(required, def.onchange), data(port && port.indexBy, name))
+            }, port);
+        }),
+    }, def));
+};
+
+function data(indexBy, name) {
+    var by = _.isFunction(indexBy) ? indexBy : _.property(indexBy);
+    return function(payload, socketIndex) {
+        var key = by(payload);
+        var state = this.state || {};
+        var index = indexBy == null ? payload :
+            _.extend(state[name] || {}, _.object([key], [payload]));
+        this.state = _.extend(state, _.object([name], [index]));
+        return key;
+    };
+}
+
+function change(required, onchange) {
+    return function(key) {
+        var inPorts = this.inPorts;
+        var state = _.mapObject(this.state, function(index, name) {
+            if (inPorts[name].options.indexBy == null) return index;
+            else return index[key];
+        });
+        var missing = _.difference(required, _.keys(state));
+        if (_.isEmpty(missing) && _.isFunction(onchange))
+            return onchange.call(this, state);
+    }
+}
--- a/components/state-component.js
+++ b/components/state-component.js
@@ -1,6 +1,7 @@
 // state-component.js
 
 var _ = require('underscore');
+var jsonpointer = require('jsonpointer');
 var promiseComponent = require('./promise-component');
 
 /**
@@ -24,9 +25,9 @@ module.exports = function(def){
 };
 
 function data(indexBy, name) {
-    var by = _.isFunction(indexBy) ? indexBy : _.property(indexBy);
+    var by = _.isFunction(indexBy) ? indexBy : pointer(indexBy);
     return function(payload, socketIndex) {
-        var key = by(payload);
+        var key = by.call(this, payload);
         var state = this.state || {};
         var index = indexBy == null ? payload :
             _.extend(state[name] || {}, _.object([key], [payload]));
@@ -46,4 +47,10 @@ function change(required, onchange) {
         if (_.isEmpty(missing) && _.isFunction(onchange))
             return onchange.call(this, state);
     }
+}
+
+function pointer(path) {
+    if (path && path.charAt(0) == '/')
+        return jsonpointer.compile(path).get;
+    else return _.property(path);
 }
--- a/components/state-component.js
+++ b/components/state-component.js
@@ -5,10 +5,19 @@ var jsonpointer = require('jsonpointer');
 var promiseComponent = require('./promise-component');
 
 /**
- * Stores the most recent packet for every inPort and calls an onchange function
- * when they change and after when all required port have received something.
- * If multiple ports both provide an indexBy property name or function,
- * only packets that both resolve to the same value with be called together
+ * Stores the last packet for every inPort and calls an onchange function
+ * when received and after when all required port have received something.
+ * If multiple ports both provide an indexBy property name, path or function,
+ * only packets that both resolve to the same key with be called together.
+ *
+ * Usage:
+ *   stateComponent({
+ *     inPorts: {} of port names to port options || [] of port names,
+ *     onchange: function(inStates, outState) {
+ *         inStates = {} of port names to state,
+ *         outState = previous result of onchange
+ *     }
+ *   });
  */
 module.exports = function(def){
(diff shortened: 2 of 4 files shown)


test/state-component-mocha.js
```diff
@@ -0,0 +1,139 @@
...
+        return component;
+    }
+    function sendData(component, port, payload) {
+        var socket = noflo.internalSocket.createSocket();
+        component.inPorts[port].attach(socket);
+        socket.send(payload);
+        socket.disconnect();
+        component.inPorts[port].detach(socket);
+    }
+    function onceData(component, resolvePort, rejectPort, sendPort, sendPayload) {
```
glenna-m: Is this function is actually used?   Did not look like it to me.  If not used, please remove.

components/state-component.js
```diff
@@ -0,0 +1,49 @@
...
+    return promiseComponent(_.defaults({
+        inPorts: _.mapObject(inPorts, function(port, name) {
+            return _.extend({
+                ondata: _.compose(change(required, def.onchange), data(port && port.indexBy, name))
+            }, port);
+        }),
+    }, def));
+};
+
+function data(indexBy, name) {
```
glenna-m: indexBy is a property in this module, but would be more flexible as a jsonpointer.

test/state-component-mocha.js
```diff
@@ -0,0 +1,139 @@
...
+    it("should reject undefined definition", function() {
+        return Promise.resolve().then(stateComponent).should.be.rejected;
+    });
+    it("should reject empty definition", function() {
+        return Promise.resolve({}).then(stateComponent).should.be.rejected;
+    });
+    it("should reject empty inPorts", function() {
+        return Promise.resolve({inPorts:{}}).then(stateComponent).should.be.rejected;
+    });
+    it("should accept array of inPort names", function() {
```
glenna-m: In most cases, the inPorts are likely to be one or more objects that look like this: 
inPorts: {
      portA: {
        datatype: 'string',
        description: "portA description",
        required: true
      },
      portB: {
        datatype: 'object',
        description: "portB description",
        required: false
      }
  }

It might be good to add a more typical test case for a configuration like this to validate correct handling of it instead of such a basic array.

components/state-component.js
catch-point: DONE

test/state-component-mocha.js
catch-point: That is not what this test is covering. Other tests both in this file and in the promise-component-mocha test that.

test/state-component-mocha.js
catch-point: DONE

test/state-component-mocha.js
glenna-m: I know that is not what this test case is covering.  I was suggesting adding a test case for this.  It is true you have a test that handles a very basic 2 port object scenario (name only).  Was suggesting it might be nice to have one that reflects how most updaters should really be calling this.  As a best practice, I think we want to encourage folks to set the datatype, description, and required flag, but there is no test in this suite that shows that.

components/promise-component.js
```diff
@@ -38,7 +38,7 @@ module.exports = function(def){
     var inPorts = _.mapObject(def.inPorts, function(port, name){
         var process = port.process || function(event, payload, socketIndex) {
             var key = 'on' + event;
-            if (port[key]) return port[key].call(this.nodeInstance, payload);
+            if (port[key]) return port[key].call(this.nodeInstance, payload, socketIndex);
```
glenna-m, line 41: Good.

components/state-component.js
```diff
@@ -0,0 +1,56 @@
...
+    var by = _.isFunction(indexBy) ? indexBy : pointer(indexBy);
```
glenna-m, line 43: What if there is no indexBy (neither function nor port) - that compiled jsonpointer will fail, no?   I don't think an indexBy is required is it?

components/state-component.js
```diff
@@ -0,0 +1,56 @@
...
+
+var _ = require('underscore');
+var jsonpointer = require('jsonpointer');
+var promiseComponent = require('./promise-component');
+
+/**
+ * Stores the most recent packet for every inPort and calls an onchange function
+ * when they change and after when all required port have received something.
+ * If multiple ports both provide an indexBy property name or function,
+ * only packets that both resolve to the same value with be called together
```
dbooth-boston: What is indexBy for?  Why is it needed?

components/state-component.js
```diff
@@ -0,0 +1,56 @@
...
+ */
+module.exports = function(def){
+    var inPorts = _.isArray(def.inPorts) ? _.object(def.inPorts, []) : def.inPorts;
```
dbooth-boston, line 26: Why would def.inPorts ever be an array?

components/state-component.js
```diff
@@ -0,0 +1,56 @@
...
+        this.state = _.extend(state, _.object([name], [index]));
+        return key;
+    };
+}
+
+function change(required, onchange) {
+    return function(key) {
+        var inPorts = this.inPorts;
+        var state = _.mapObject(this.state, function(index, name) {
+            if (inPorts[name].options.indexBy == null) return index;
```
dbooth-boston: What is this index stuff about?

components/state-component.js
catch-point, line 43: if indexBy is undefined, 'by' function(obj) will equivalent to obj.undefined, but never used with the most recent changes.

components/state-component.js
catch-point: This provides a hook for vni management. i.e. a function that will extract the vnid from the payload.

components/state-component.js
catch-point, line 26: If the wrapper is not provided with any metadata, only a function. It would be easier to allow it to provide a list of port names (based on the function argument names).

dbooth-boston, line 26: ok, I like that

components/state-component.js
catch-point: The main goal of this is to take payloads from different ports and map them together based on some common key (such as an vnid). To do that this code indexes the incoming payload by a key (provided by the caller) and then calls the onchange function with payloads that have a matching key.

components/state-component.js
```diff
@@ -0,0 +1,88 @@
...
+        this.inPayloads = _.extend(payloads, _.object([name], [index]));
+        return key;
+    };
+}
+
+function fireChangeEvent(required, onchange) {
+    return function(key) {
+        var wild = _.pick(this.inPayloads, isWild.bind(this));
+        if (_.isUndefined(key)) {
+            // no key for this event, first with only wild matching payloas
```
glenna-m: Nit - minor spelling error - payloas should be payload.

components/state-component.js
```diff
@@ -0,0 +1,88 @@
...
+            var missing = _.difference(required, _.keys(wild));
+            if (_.isEmpty(missing) && _.isFunction(onchange)) {
+                var self = this;
+                return Promise.resolve(onchange.call(this, wild, this.outPayload)).then(function(result){
+                    return self.outPayload = result;
+                });
```
glenna-m, line 65: shouldn't there be error handling if it's not wild and there is no matching index?

components/state-component.js
catch-point: FIXED

components/state-component.js
```diff
@@ -0,0 +1,89 @@
...
+ * Usage:
+ *   stateComponent({
+ *     inPorts: {name:{indexBy:function(payload)}} of port names to port options || [] of port names,
+ *     onchange: function(inPayloads, outPayload) {
+ *         inPayloads = {name:payload} of port names to last payload,
+ *         outPayload = previous result of onchange
+ *     }
+ *   });
+ */
+module.exports = function(def){
```
glenna-m, line 24: @dbooth-boston  asked me to use nodeDef instead of def in javascript-wrapper.  That feedback probably applies here too.

components/state-component.js
```diff
@@ -0,0 +1,89 @@
...
+                ondata: _.compose(
+                    fireChangeEvent(required, def.onchange),
+                    indexPayload(port && port.indexBy, name)
+                )
+            }, port);
+        })
+    }, def));
+};
+
+function indexPayload(indexBy, name) {
```
glenna-m, line 42: @dbooth-boston has been asking me to include a comment above each function describing what it does, parameters, and return.

glenna-m, line 42: suggest renaming name to portName for clarity

components/state-component.js
catch-point, line 24: I'm not sure it would necessarily apply as the definition changes from "node" to "component" as it moves down the stack. I'm also not sure adding the term "node" adds clarity to what it is (as the term is generally overloaded).

components/state-component.js
```diff
@@ -0,0 +1,89 @@
...
+ *     onchange: function(inPayloads, outPayload) {
+ *         inPayloads = {name:payload} of port names to last payload,
+ *         outPayload = previous result of onchange
+ *     }
+ *   });
+ */
+module.exports = function(def){
+    if (!_.isFunction(def.onchange)) throw Error("onchange must be a function");
+    var inPorts = _.isArray(def.inPorts) ? _.object(def.inPorts, []) : def.inPorts;
+    var required = _.compact(_.map(inPorts, function(port, name){
```
dbooth-boston, line 27: I think we should treat the 'required' attribute (on an inport) as meaning: "this port requires a connection -- either an IIP or an edge connection to another port".  It should _not_ mean: "this port requires data to have arrived before firing the updater".  _Every_ connected port must have data before firing the updater, regardless of whether the port has the 'required' attribute set.  Is this how the code below works?

components/state-component.js
```diff
@@ -0,0 +1,89 @@
...
+};
+
+function indexPayload(indexBy, name) {
+    var by = _.isFunction(indexBy) ? indexBy : pointer(indexBy);
+    return function(payload, socketIndex) {
+        // determine a key string or undefined if no key should be used
+        var key = indexBy == null ? undefined : ('' + by.call(this, payload));
+        var payloads = this.inPayloads || {};
+        var index = _.isUndefined(key) ? payload :
+            _.extend(payloads[name] || {}, _.object([key], [payload]));
```
glenna-m, line 49: Are you expecting to have an array of keys and payloads in your _.object call here?  Since you only have the one payload, I'm thinking probably not?  If so, wouldn't { [key] : [payload] } be more clear?  It rules out the _.object expecting arrays

catch-point, line 49: What do you mean? Can you show it in code?

components/state-component.js
glenna-m, line 24: @dbooth-boston the def here is literally being passed from the wrapper to state-component with just the addition of of the onchange updater hook.  If one is a nodeDef, the other one is.

components/state-component.js
```diff
@@ -0,0 +1,89 @@
...
+            if (_.isEmpty(missing)) {
+                var self = this;
+                return Promise.resolve(onchange.call(this, wild, this.outPayload)).then(function(result){
+                    return self.outPayload = result;
+                });
+            }
+        } else {
+            // a particular key must match payload (or be wild matching)
+            var payloads = _.extend(_.mapObject(this.inPayloads, _.property(key)), wild);
+            var missing = _.difference(required, _.keys(payloads));
```
catch-point, line 70: I'll change this to check both required and attached ports. Furthermore, I'll change this to throw an error if a required port is not attached.

components/state-component.js
```diff
@@ -0,0 +1,89 @@
...
+    }
+}
+
+function pointer(path) {
+    if (path && path.charAt(0) == '/')
+        return jsonpointer.compile(path).get;
+    else return _.property(path);
+}
+
+function isWild(data, name) {
```
glenna-m, line 87: Define what isWild really means.  I know from David's slides, but someone reading code won't.

glenna-m:
The name state-component implies to someone from the noflo-world that this is a component that can be configured and executed as is from the gui - clearly not what this is intended for.  Is there a better name we could use to avoid causing noflo confusion?

components/state-component.js
glenna-m, line 87: You are not currently using the data parameter in this function.  Is there a reason you need to pass it?   If so, please document it.   Or since it is not used, perhaps it should be removed?

test/state-component-mocha.js
```diff
@@ -0,0 +1,288 @@
...
+            // have the handler call a Promise resolve function to
+            // check that the data sent on the in port is passed to the handler
+            return new Promise(function(callback){
+                handler = callback;
+                sendData(component, 'a', {aid:"1", c: "A"});
+                sendData(component, 'a', {aid:"2", c: "A"});
+                sendData(component, 'a', {aid:"3", c: "A"});
+                sendData(component, 'b', {bid:"2", c: "B"});
+            });
+        }).should.become({a: {aid: "2", c: "A"}, b: {bid: "2", c: "B"}});
```
glenna-m, line 194: I'm curious what the use case is for this?  @jamesrdf  @dbooth - Do we have an RDF reason why we would want to have data on port b with a different index, overriding the index value on port a?

test/state-component-mocha.js
```diff
@@ -0,0 +1,288 @@
...
+                    handler = callback;
+                    sendData(component, 'c', {d:"C1"});
+                });
+            }).then(function(){
+                return new Promise(function(callback){
+                    handler = callback;
+                    sendData(component, 'c', {d:"C2"});
+                });
+            });
+        }).should.become(['c']); // a should be absent because it's indexed
```
glenna-m: This test is confusing to me.  Are you basically trying to verify that you got the correct last port?

test/state-component-mocha.js
```diff
@@ -0,0 +1,288 @@
...
+                handler = callback;
+                sendData(component, 'property', 'id');
+                sendData(component, 'a', {id:"1", c: "A"});
+                sendData(component, 'a', {id:"2", c: "A"});
+                sendData(component, 'a', {id:"3", c: "A"});
+                sendData(component, 'b', {id:"2", c: "B"});
+            });
+        }).should.become({a: {id: "2", c: "A"}, b: {id: "2", c: "B"}});
+    });
+    function createComponent(getComponent) {
```
glenna-m, line 269: createComponent and sendData are helper methods with identical code copied and pasted between promise-component-mocha.js and state-component-mocha.js.  They should be a in a common helper file that both can use so we don't proliferate copy/pasted code and have one place to maintain them.

test/state-component-mocha.js
catch-point, line 194: There is no overriding here. The payload, identified by "2", from port a is matched with the payload, identified by "2", from port b and both are passed together to the onchange handler.

test/state-component-mocha.js
catch-point, line 126: Testing an edge case to ensure an unindexed payload is not called with an indexed payload.

test/state-component-mocha.js
glenna-m, line 194: OK.  Got it.  This makes sense now.   I think one way to make this more intuitive to others would be to change the test "it" wording - perhaps something like

"should execute onchange with matched incoming data packets using index key"

And include a comment about how packets can come into two ports in any order and you need to match the right ones together for onchange.

test/state-component-mocha.js
```diff
@@ -0,0 +1,288 @@
...
+            return new Promise(function(callback){
+                handler = callback;
+                sendData(component, 'a', {aid:"1", c: "A"});
+                sendData(component, 'a', {aid:"2", c: "A"});
+                sendData(component, 'a', {aid:"3", c: "A"});
+                sendData(component, 'b', {bid:"2", c: "B"});
+            });
+        }).should.become({a: {aid: "2", c: "A"}, b: {bid: "2", c: "B"}});
+    });
+    it("map indexed data by pointer", function() {
```
glenna-m, line 196: How about something like "map indexed data by jsonpointer" ?   That lets someone new to jsonpointers know how to look it up.

test/state-component-mocha.js
```diff
@@ -0,0 +1,288 @@
...
+                sendData(component, 'hello', "world");
+            }).then(function(){
+                return new Promise(function(callback){
+                    handler = callback;
+                    sendData(component, 'hello', "again");
+                });
+            });
+        }).should.become("world");
+    });
+    it("shouldn't confuse an undefined key with non-indexed port", function() {
```
glenna-m, line 75: How about something like

shouldn't confuse data with an undefined key on indexed port with incoming data on a non-indexed port

And add something somewhere that explains the difference between an indexed and wildcarded port.

dbooth-boston:
I now have a little better understanding of what this is doing, and there are two things that make me uncomfortable.  One is the extent to which Promises are being used in general, both in the state-component and the underlying promise-component.  I am convinced that Promises are the right tool for the job of handling asynchronous (long-running) updaters, but I think they're being introduced too low down in our framework and therefore used more heavily than they need to me, and that makes the code a lot harder for me to follow than it otherwise would be.  (See http://dbooth.org/2016/jspromise.pdf .)  I would rather that Promises only be used at the point where they are needed, such as when a potentially long-running updater is to be called. Can we redesign to do that?

The second thing more specifically about state-component that I'm uncomfortable about is that the logic for remembering payloads is being done separately from the logic for remembering input states, and that also makes it harder for me to follow what's going on.  The use of indexBy also seems more general than I think it needs to be, because we only need to look for a 'vnid' property specifically.  If you compare this logic to the selfReceiveEvent function starting at line 129 in https://github.com/rdf-pipeline/noflo-rdf-components/blob/dbooth-experiments/components/Map.js I think the logic in selfReceiveEvent is much easier to follow, and that function can get registered _directly_ on the NoFlo component, without the need for other layers of event dispatching.  Granted, selfReceiveEvent would need to be modified slightly to enable updaters to be asynchronous, but that could be done by moving the call to maybeSendEvent from line 168 into the maybeFireUpdater function, and using a Promise inside maybeFireUpdater.

In short, I think I would rather give up some generality in favor of making the code a lot easier for folks like me to follow.

PR #10, opened 2016-01-27
Javascript wrapper

glenna-m:
Ready for review with James' state-component

### DIFF
--- a/components/javascript-wrapper.js
+++ b/components/javascript-wrapper.js
@@ -1,4 +1,3 @@
-
 /**
  * This file contacts the common code used by all of the variations of any RDF
  * node componentin the noflo implementation of the RDF Pipeline.
@@ -12,67 +11,82 @@ var stateComponent = require('./state-component');
  * code to ensure updater callback, and then invoking the RDF pipeline framework to complete building
  * the component and setting up the vni and state metadata.
  *
- * @param def updater definition. This can be a simple updater function, or an object that contains 
+ * @param nodeDef updater node definition. This can be a simple updater function, or an object that contains 
  *        updater metadata (e.g., input/output port definitions, description, icons), and an updater.
  *
  * @return a promise to create the noflo rdf component
  */
-module.exports = function(def) { 
+module.exports = function(nodeDef) { 
 
+    // Save the caller's node instance 
     var self = this;
+    
+    var updaterArgs = null;
+    if ( _.isUndefined(nodeDef) || 
+       ( ! _.isFunction(nodeDef) && ( _.isEmpty( nodeDef ) || ( _.isEmpty( nodeDef.inPorts ))))) {
+       // Got nothing to work with - no input ports and no updater, so create a default updater and input port 
+       nodeDef = { 
+           inPorts: {
+               input: {
+                   datatype: 'object',
+                   description: "default input port" 
+               }
+           },
+           updater: defaultUpdater
+       };
+    }  else if ( _.isUndefined( nodeDef.updater ) && !_.isFunction(nodeDef)) { 
+       // Got some node metadata, but no node updater - use the metadata with default updater
+       nodeDef = _.extend( nodeDef, 
+                           { updater: defaultUpdater } );
+    } else if ( _.isFunction( nodeDef ) ) { 
+        // Got an updater but no metadata - use updater args as input port list
+        // This assumes that all updater arguments are input
+        updaterArgs = introspect( nodeDef );
+        nodeDef = { inPorts: updaterArgs,
+                    updater: nodeDef };
+    } 
 
-    if ( _.isUndefined(def) || ( _.isEmpty(def) && ! _.isFunction( def ))) { 
-        // Nothing to work with - no metadata or updater - can't proceed.
-        throw new Error('javascript-wrapper found no updater definition');
-    }
-
-    if ( _.isUndefined( def.updater ) && !_.isFunction(def)) { 
-       // got called with def metadata, but no updater for this one
-        return stateComponent( def );
-    }
-     
-    // Have an updater - figure out the updater arguments 
-    var updaterArgs = ( _.isFunction(def) ) ? introspect( def ) : introspect( def.updater );
-
-    // Do we have any input port definition?  If we were only passed an updater function, we might not.
-    // In that case, build the list of input port names from the updater function arguments.  We will 
-    // assume that all arguments are input ports.
-    if ( _.isFunction( def ) ) { 
-        def = { updater: def, inPorts: updaterArgs };
-    }
-
-    // set up an onchange callback so we can invoke the updater when we have changes to process
-    var updaterCallback = function() { 
-        return function( state ) {
-
-            // Do we have any updaterArgs?  We might not if we do not have source code
-            if ( _.isEmpty( updaterArgs ) ) { 
-               // No updaterArgs, so just pass the state
-               def.updater.call( self, state );
-
-            } else {
-                // Have one or more parameters - get them in the right order and pass them 
-                var parameters = _.map( updaterArgs, function( arg ) { 
-                    var parameter = state[arg]; 
-                    if ( _.isUndefined(parameter) && def.outPorts && def.outPorts.hasOwnProperty( arg ) ) {
-                        // passing in the name of the output port - is this correct? 
-                        parameter = arg;
-                    }
-                    return parameter; 
-                });
-
-                def.updater.apply( self, parameters );
-            }
-        }
-    }
+    // Have an updater now, no matter what, be it default or not - figure out what the updater arguments are
+    var updaterArgs = (_.isNull( updaterArgs )) ? introspect( nodeDef.updater ) : updaterArgs; 
 
     return stateComponent( 
         _.defaults({
-            onchange: updaterCallback()
-        }, def)
+            // set up an onchange callback so we can invoke the updater when we have changes to process
+            onchange: 
+// function() { 
+                // return function( state, payload ) {
+                function( state, payload ) {
+                    // Do we have any updaterArgs?  We might not if we do not have source code
+                    if ( _.isEmpty( updaterArgs ) ) { 
+                       // No updaterArgs, so just pass the state
+                       return nodeDef.updater.call( self, state, payload );
+        
+                    } else {
+                        // Have one or more parameters - get them in the right order and pass them 
+                        var parameters = _.map( updaterArgs, function( arg ) { 
+                            var parameter = state[arg]; 
+                            if ( _.isUndefined(parameter) && nodeDef.outPorts && nodeDef.outPorts.hasOwnProperty( arg ) ) {
+                                // passing in the name of the output port - is this correct? 
+                                parameter = arg;
+                            }
+                            return parameter; 
+                        });
+        
+                        return nodeDef.updater.apply( self, parameters );
+                    }
+                }
+            // }
+        }, nodeDef)
       ); 
 
 }; // module.exports
+
+// Define a default updater stub function to be used if the caller 
+// did not pass one in.  Simply returns whatever the original input was so it
+// can be sent through the output port to the next component
+var defaultUpdater = function() { 
+    return { out: arguments };
+}
 
 // Introspect the updater to determine what its parameters are.
 // The algorithm here comes from the following sources:
--- a/components/javascript-wrapper.js
+++ b/components/javascript-wrapper.js
@@ -1,0 +1,111 @@
+
+/**
+ * This file contacts the common code used by all of the variations of any RDF
+ * node componentin the noflo implementation of the RDF Pipeline.
+ */
+
+var _ = require('underscore');
+var stateComponent = require('./state-component');
+
+/**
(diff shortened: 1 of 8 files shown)


components/javascript-wrapper.js
```diff
@@ -0,0 +1,111 @@
...
+
+var _ = require('underscore');
+var stateComponent = require('./state-component');
+
+/**
+ * This module provides a Javascript wrapper for updaters.  It is responsible for setting up the
+ * code to ensure updater callback, and then invoking the RDF pipeline framework to complete building
+ * the component and setting up the vni and state metadata.
+ *
+ * @param def updater definition. This can be a simple updater function, or an object that contains
```
dbooth-boston: Suggest renaming def -> updaterDef

components/javascript-wrapper.js
```diff
@@ -0,0 +1,111 @@
...
+module.exports = function(def) { 
+
+    var self = this;
+
+    if ( _.isUndefined(def) || ( _.isEmpty(def) && ! _.isFunction( def ))) { 
+        // Nothing to work with - no metadata or updater - can't proceed.
+        throw new Error('javascript-wrapper found no updater definition');
+    }
+
+    if ( _.isUndefined( def.updater ) && !_.isFunction(def)) {
```
dbooth-boston: if (!def.updater) def.updater = defaultJavascriptUpdater;
if (!_.isFunction(def)) throw FatalError( ... )   -- this is a misconfiguration of the pipeline and does not need to be recoverable

components/javascript-wrapper.js
```diff
@@ -0,0 +1,111 @@
...
+
+    // Do we have any input port definition?  If we were only passed an updater function, we might not.
+    // In that case, build the list of input port names from the updater function arguments.  We will 
+    // assume that all arguments are input ports.
+    if ( _.isFunction( def ) ) { 
+        def = { updater: def, inPorts: updaterArgs };
+    }
+
+    // set up an onchange callback so we can invoke the updater when we have changes to process
+    var updaterCallback = function() {
```
catch-point: This doesn't need to be a higher order function

components/javascript-wrapper.js
```diff
@@ -0,0 +1,111 @@
...
+    // Do we have any input port definition?  If we were only passed an updater function, we might not.
+    // In that case, build the list of input port names from the updater function arguments.  We will 
+    // assume that all arguments are input ports.
+    if ( _.isFunction( def ) ) { 
+        def = { updater: def, inPorts: updaterArgs };
+    }
+
+    // set up an onchange callback so we can invoke the updater when we have changes to process
+    var updaterCallback = function() { 
+        return function( state ) {
```
catch-point: The latest version of state-component provides the previous result as a second argument. Use that for the outPort state.

components/javascript-wrapper.js
```diff
@@ -0,0 +1,111 @@
...
+                var parameters = _.map( updaterArgs, function( arg ) { 
+                    var parameter = state[arg]; 
+                    if ( _.isUndefined(parameter) && def.outPorts && def.outPorts.hasOwnProperty( arg ) ) {
+                        // passing in the name of the output port - is this correct? 
+                        parameter = arg;
+                    }
+                    return parameter; 
+                });
+
+                def.updater.apply( self, parameters );
```
catch-point: The onchange function should return something to send to the outPort

components/javascript-wrapper.js
```diff
@@ -0,0 +1,111 @@
...
+ *
+ * @return a promise to create the noflo rdf component
+ */
+module.exports = function(def) { 
+
+    var self = this;
+
+    if ( _.isUndefined(def) || ( _.isEmpty(def) && ! _.isFunction( def ))) { 
+        // Nothing to work with - no metadata or updater - can't proceed.
+        throw new Error('javascript-wrapper found no updater definition');
```
dbooth-boston: Let's provide a default input port and a defaultJavascriptUpdater if nothing was specified.  That will make it easy for people to create a node as a stub and then modify it later.  The defaultJavascriptUpdater would just pass the input to the output.  BTW, by default an input port should not allow multiple connections.

components/javascript-wrapper.js
```diff
@@ -0,0 +1,111 @@
...
+    }
+
+    if ( _.isUndefined( def.updater ) && !_.isFunction(def)) { 
+       // got called with def metadata, but no updater for this one
+        return stateComponent( def );
```
dbooth-boston: In that case:
def.updater = defaultJavascriptUpdater;

components/javascript-wrapper.js
```diff
@@ -0,0 +1,111 @@
...
+                    return parameter; 
+                });
+
+                def.updater.apply( self, parameters );
+            }
+        }
+    }
+
+    return stateComponent( 
+        _.defaults({
```
catch-point, line 53: It might be nice to use _.extend here to allow the caller to override the onchange function (bypassing the features of this wrapper).

test/javascript-wrapper-mocha.js
```diff
@@ -0,0 +1,203 @@
...
+                in: { datatype: 'string', 
+                      description: "input description",
+                      required: true }
+            }
+        }).then(jswrapper).then(commonTest.createComponent).then(function(component){
+            return _.keys(component.inPorts);
+        }).should.eventually.contain('in');
+    });
+
+    it("should trigger onchange function if no updater", function() {
```
catch-point: What should happen if passed both an onchange and updater function?

test/javascript-wrapper-mocha.js
```diff
@@ -0,0 +1,203 @@
...
+                handler = callback;
+                commonTest.sendData(component, 'in1', "alpha and ");
+                commonTest.sendData(component, 'in2', "omega");
+            });
+        }).should.become( 'alpha and omega' );
+    });
+
+  // This introspect tests must be run in test mode with NODE_ENV=test 
+  // e.g., NODE_ENV=test mocha since we are testing a private function
+  describe('#introspect - run with NODE_ENV=test to fully execute these tests', function() {
```
catch-point, line 189: Might be better off pulling out the introspect function into its own file that can more easily be tested in isolation.

test/javascript-wrapper-mocha.js
```diff
@@ -0,0 +1,203 @@
...
+
+var sinon = require('sinon');
+
+var _ = require('underscore');
+var noflo = require('noflo');
+
+var jswrapper = require('../components/javascript-wrapper');
+var commonTest = require('./common-test');
+
+describe('javascript-wrapper', function() {
```
catch-point, line 22: Can you add a test to cover the return state of an updater?

components/javascript-wrapper.js
```diff
@@ -0,0 +1,111 @@
...
+ * the component and setting up the vni and state metadata.
+ *
+ * @param def updater definition. This can be a simple updater function, or an object that contains 
+ *        updater metadata (e.g., input/output port definitions, description, icons), and an updater.
+ *
+ * @return a promise to create the noflo rdf component
+ */
+module.exports = function(def) { 
+
+    var self = this;
```
dbooth-boston, line 22: Please add a comment saying what 'this' is expected to be bound to.  A VNI?  A NoFlo node instance?  A NoFlo port?  A component factory?

components/javascript-wrapper.js
```diff
@@ -0,0 +1,111 @@
...
+    }
+
+    // set up an onchange callback so we can invoke the updater when we have changes to process
+    var updaterCallback = function() { 
+        return function( state ) {
+
+            // Do we have any updaterArgs?  We might not if we do not have source code
+            if ( _.isEmpty( updaterArgs ) ) { 
+               // No updaterArgs, so just pass the state
+               def.updater.call( self, state );
```
dbooth-boston: What is 'state'? Is it the node's current output state object?  We also need to distinguish in our naming between a State object {data: "jane", lm: "..."} and the user-controlled data within that state, i.e., state.data .  We could reconsider the names of these if anyone thinks there would be better names.

components/javascript-wrapper.js
```diff
@@ -0,0 +1,111 @@
...
+
+            } else {
+                // Have one or more parameters - get them in the right order and pass them 
+                var parameters = _.map( updaterArgs, function( arg ) { 
+                    var parameter = state[arg];
```
dbooth-boston: This should be 'inputs' -- not 'state' -- shouldn't it?  And it's more complex than this because only the state.data objects should be passed, and 'inputs' is a hash of hashes because of the possibility of multiple connections to the same inport (what NoFlo calls 'sockets'), and because if an inport has the allowMultipleConnections option set then the parameter should receive a hash of state.data objects, whereas if !allowMultipleConnections then it should receive a state.data object directly.  See slide 17 ("") of http://dbooth.org/2013/pipeline/RDF-Pipeline-Implementation.pdf .

test/javascript-wrapper-mocha.js
catch-point, line 22: Can you add a test to ensure the updater's this context is called with the nodeInstance?

components/javascript-wrapper.js
catch-point: The updater this context should be the nodeInstance

components/javascript-wrapper.js
glenna-m: The callback from the state-component is it's view of state right now.   The state being received here is currently a hash object with input port name as index into to the port payload. 

When we have a true inputs object, this could change.

components/javascript-wrapper.js
glenna-m, line 53: What would you extend it with?   If you don't want this wrapper, you can call a different wrapper, or call state-component (or vni-component) directly, no?

components/javascript-wrapper.js
glenna-m: Will do for next check  in

components/javascript-wrapper.js
glenna-m, line 22: In the context here, this is the updater context.  Will add the comment.

components/javascript-wrapper.js
glenna-m: OK - will add the defaults for the next check in.

The javascript-wrapper does not know how many connections we have - it only has the port definitions which are not even instantiated yet at the time we see it - that policy enforcement needs to be lower.

components/javascript-wrapper.js
glenna-m: I think we need to sort out what that state should be.  IMHO, James's port/payload hash (which is what that state is now) , should possibly become your state object with the lm.  It has the data required to build it.  Will check in some ideas on that later, perhaps.

components/javascript-wrapper.js
glenna-m: Will fix this.

test/javascript-wrapper-mocha.js
glenna-m, line 22: Added a context test, and modified "should trigger default updater function if no updater" to verify default updater return state.

test/javascript-wrapper-mocha.js
glenna-m: What's the use case for allowing that?

test/javascript-wrapper-mocha.js
glenna-m, line 189: I would if I thought it would be more widely used.  This is a custom-to-javascript introspector though, so I don't see other wrappers as likely to use it.  Also, it's trivial to simply do 
NODE_ENV=test mocha 
to run the test suite.   That is the default way that many teams have their unit testing configured to execute anyway.

components/javascript-wrapper.js
glenna-m: Will do.

components/javascript-wrapper.js
dbooth-boston: Actually nodeDef might be better

components/javascript-wrapper.js
dbooth-boston: I'm surprised to see 'def' still used when the updater is called, because I think of def as only being used during graph creation -- not at runtime when the pipeline is sending events and processing data.  I think of it as only a compiletime structure.

Also, the updater needs a way to indicate whether or not it actually modified the node's state, so that the framework can know whether to send events downstream if it did.  The API for this was shown on slides 5-9 of http://dbooth.org/2013/pipeline/RDF-Pipeline-Implementation.pdf , but if we switch to a one-state-per-updater model then we may want to change this API.  But somehow the updater needs to be able to indicate that it did or did not end up changing the state.

components/javascript-wrapper.js
```diff
@@ -0,0 +1,125 @@
...
+        // This assumes that all updater arguments are input
+        updaterArgs = introspect( nodeDef );
+        nodeDef = { inPorts: updaterArgs,
+                    updater: nodeDef };
+    } 
+
+    // Have an updater now, no matter what, be it default or not - figure out what the updater arguments are
+    var updaterArgs = (_.isNull( updaterArgs )) ? introspect( nodeDef.updater ) : updaterArgs; 
+
+    return stateComponent(
```
catch-point, line 52: Switch to using vniComponent

PR #11, opened 2016-01-27
Add vni-component

catch-point:


glenna-m:
Do you still want review on this, James, or are you just keeping it to refer to as you do integration work?

PR #12, opened 2016-01-28
Javascript implementation of lm.

glenna-m:
A javascript implementation of a more robust version of lm.


dbooth-boston:
Did you see the code for newLm() in https://github.com/rdf-pipeline/noflo-rdf-components/blob/dbooth-experiments/components/Map.js lines 572-613?  That function represents an LM as a fixed-width string starting with "LM", to (nearly) match the perl implementation, be easily recognizable in logs, allow easy comparison, allow it to be used directly in an HTTP header, and allow it to be round-trip serializable/deserializable.  The version that you propose is simpler, but it does not meet these additional objectives, which I think are helpful.  LMK if you think differently.

glenna-m:
Yes, I saw it, but it felt more like perl transcribed to Javascript more than anything else.    In particular, I was not comfortable with the rounding and forcing it to a string in order to compare.    I can pretty easily address the logging and serialization issues.  I'll post an update for that shortly.

dbooth-boston:
What rounding?   

The main benefit I see of using a string instead of an object is that it allows the same representation to be used everywhere.  What is the advantage that you see in using an object instead?

dbooth-boston:
Oh, on the rounding maybe you mean the rounding to milliseconds.  The EC 5 spec says that the precision of a Date object is only milliseconds, so AFAICT we should be rounding to milliseconds.

Strings can be compared directly using ===, which we could not easily do if we used objects and allowed serialization/deserialization.

Bottom line: I just thought it would make them overall easier to work with if they were consistently strings everywhere.  But if you and James both think there would be greater benefit in having them as objects I guess it's okay with me.

components/lm.js
```diff
@@ -0,0 +1,85 @@
...
+ *
+ * @param string optional string LM
+ *
+ * @return new LM is no lm string or a deserialized Lm built from the 
+ *         string if a good Lm string was given
+ */
+function Lm( string ) { 
+
+   // If we get an LM string, parse & validate it
+   var components = (string) ? arguments[0].match(/LM(\d+)\.(\d+)/) : null;
```
dbooth-boston: anchor, but allow spaces before/after

dbooth-boston:
Rename to LmFactory, then LGTM to merge.

PR #14, opened 2016-02-01
Add event-component-factory

catch-point:
This gives the outPorts the same events as inPorts, which will allow both ports to be used the same way, regardless of the direction of the underlying socket. It will also give us a common way to register for event handlers, where they are noflo events or custom events, without having to pass around callback functions.

### DIFF
--- a/components/event-component-factory.js
+++ b/components/event-component-factory.js
@@ -1,0 +1,114 @@
+// component-factory.js
+
+var _ = require('underscore');
+var noflo = require('noflo');
+
+/**
+ * Creates a noflo Component factory function from a component definitation,
+ * registers any event handlers on definitation. Triggers ondata events for outPorts.
+ *
+ * Usage:
+ *  componentFactory({
+ *      description: "text",
+ *      icon: 'fontawesome-icon-name',
+ *      inPorts: {
+ *          portName: {
+ *              ondata: function(payload)
+ *          }
+ *      },
+ *      outPorts: {
+ *          portName: {
+ *              ondata: function(payload)
+ *          }
+ *      },
+ *      onicon: function(icon)
+ *   });
+ */
+module.exports = function(def){
+    if (!def) throw Error("No parameter");
+    return function() {
+        // noflo requires each port and nodeInstance to have its own options object
+        var nodeInstance = new noflo.Component({
+            outPorts: _.mapObject(def.outPorts, _.clone),
+            inPorts: _.mapObject(def.inPorts, _.clone)
+        });
+        triggerPortEvents(nodeInstance.outPorts);
+        registerPorts(nodeInstance.outPorts, def.outPorts);
+        registerPorts(nodeInstance.inPorts, def.inPorts);
+        registerListeners(nodeInstance, def);
+        nodeInstance.description = def.description;
+        nodeInstance.setIcon(def.icon);
+        return nodeInstance;
+    };
+};
+
+/**
+ * Fires events when OutPort functions are called on port.
+ */
+function triggerPortEvents(ports) {
+    _.each(ports, function(port) {
+        // attach and detach already trigger events
+        port.connect = _.partial(thenTrigger, port.connect, 'connect');
+        port.beginGroup = _.partial(thenTrigger, port.beginGroup, 'begingroup');
+        port.send = _.partial(thenTrigger, port.send, 'data');
+        port.endGroup = _.partial(thenTrigger, port.endGroup, 'endgroup');
+        port.disconnect = _.partial(thenTrigger, port.disconnect, 'disconnect');
+    });
+}
+
+/**
+ * Emit event after function is called
+ * @param fn function to call
+ * @param event event to emit
+ */
+function thenTrigger(fn, event /* arguments to fn and event */) {
+    var args = _.toArray(arguments).slice(2);
+    var ret = fn.apply(this, args);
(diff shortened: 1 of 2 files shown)


components/event-component-factory.js
```diff
@@ -0,0 +1,109 @@
+// component-factory.js
+
+var _ = require('underscore');
+var noflo = require('noflo');
+
+/**
+ * Creates a noflo Component factory function from a component definitation,
+ * registers any event handlers on definitation. Triggers ondata events for outPorts.
```
dbooth-boston, line 8: If I understand correctly, the purpose of triggering ondata events for outPorts is that it might provide a convenient mechanism for the Framework to invoke other functions in the framework.  But since I have not yet seen how the Framework would benefit (instead of calling another function directly), I do not yet feel convinced that we should keep it.  We can wait a bit and see though.

components/event-component-factory.js
```diff
@@ -0,0 +1,109 @@
...
+        port.send = _.partial(thenTrigger, port.send, 'data');
+    });
+}
+
+/**
+ * Emit event after function is called
+ * @param fn function to call
+ * @param event event to emit
+ */
+function thenTrigger(fn, event /* arguments to fn and event */) {
```
dbooth-boston, line 62: This is kind of cool.  Am I correct in understanding that this is used to permit a series of functions to be registered on a port, to be all called when port.send(...) is invoked?

And were you thinking that this would mostly be used for testing, or also for the Framework implementation?

components/event-component-factory.js
```diff
@@ -0,0 +1,109 @@
...
+ * @param name starts with 'on' if result is true
+ */
+function isListener(fn, name) {
+    return _.isFunction(fn) && name.indexOf('on') === 0;
+}
+
+/**
+ * Registers the function handler with the event type for this EventEmitter.
+ * @param fn function handler
+ * @param name event type prefixed with 'on'
```
dbooth-boston, line 109: Please add @this documentation to each function definition comment:
https://en.wikipedia.org/wiki/JSDoc#JSDoc_tags

components/event-component-factory.js
```diff
@@ -0,0 +1,109 @@
+// component-factory.js
```
dbooth-boston, line 1: Overall comment: nice coding style.

components/event-component-factory.js
```diff
@@ -0,0 +1,109 @@
...
+    return _.isFunction(fn) && name.indexOf('on') === 0;
+}
+
+/**
+ * Registers the function handler with the event type for this EventEmitter.
+ * @param fn function handler
+ * @param name event type prefixed with 'on'
+ */
+function register(fn, name) {
+    this.on(name.substring(2), fn);
```
dbooth-boston, line 112: I _think_ this means that string manipulation will not be used in dispatching event handling, correct?  It looks like it is only being used a component (or component factory) creation time, right?

components/event-component-factory.js
catch-point, line 62: This is to fire the 'data' event when the send function is called. This is to allow code (that will track output payloads) to register for data events on the output port.

components/event-component-factory.js
catch-point, line 112: This register function is only call from within a component factory function call.

test/event-component-factory-mocha.js
```diff
@@ -0,0 +1,84 @@
...
+    it("should reject undefined definition", function() {
+        return Promise.resolve().then(componentFactory).should.be.rejected;
+    });
+    it("should trigger in port ondata function", function() {
+        var handler;
+        return Promise.resolve({
+            inPorts:{
+                'input':{
+                    ondata: function(payload) {
+                        handler(payload);
```
dbooth-boston, line 22: Hard to figure out what 'handler' will eventually be.  Why wasn't it set directly on line 17?

test/event-component-factory-mocha.js
```diff
@@ -0,0 +1,84 @@
...
+                    }
+                }
+            }
+        }).then(componentFactory).then(createComponent).then(function(component){
+            // have the handler call a Promise resolve function to
+            // check that the data sent on the port is passed to the handler
+            return new Promise(function(callback){
```
dbooth-boston, line 29: Where is the actual parameter for this function?   I'm trying to figure out what function 'callback' will be.

components/event-component-factory.js
dbooth-boston, line 112: ok, thanks

dbooth-boston:
Subject to the above comments, this LGTM to merge.  I like it.

test/event-component-factory-mocha.js
catch-point, line 22: handler will resolve a later promise to complete the test.

test/event-component-factory-mocha.js
catch-point, line 29: callback is the promise function that will resolve it. See http://www.html5rocks.com/en/tutorials/es6/promises/#toc-api

components/event-component-factory.js
```diff
@@ -0,0 +1,113 @@
...
+ *      },
+ *      outPorts: {
+ *          portName: {
+ *              ondata: function(payload)
+ *          }
+ *      },
+ *      onicon: function(icon)
+ *   });
+ */
+module.exports = function(def){
```
dbooth-boston, line 27: def -> nodeDef

PR #16, opened 2016-02-01
Add merge-patient-lab-iips-mocha.js

catch-point:
Add a stub merge-patient-lab-iips test for sprint 5

### DIFF
--- a/test/merge-patient-lab-iips-mocha.js
+++ b/test/merge-patient-lab-iips-mocha.js
@@ -7,91 +7,90 @@ chai.use(chaiAsPromised);
 
 var _ = require('underscore');
 var noflo = require('noflo');
-var componentFactory = require('../components/merge-patient-lab-iips.js');
+var getComponent = require('../components/merge-patient-lab-iips').getComponent;
 
 describe('merge-patient-lab-iips', function() {
-    var component = createComponent(componentFactory);
-    it("should have rdfpf object on nodeInstance", function() {
-        return Promise.resolve(component).then(function(component){
-            return component.rdfpf;
+    it("should have rpf object on nodeInstance", function() {
+        return Promise.resolve(createComponent()).then(function(component){
+            return component.rpf;
         }).should.be.an('object');
     });
     it("should have createVni function", function() {
-        return Promise.resolve(component).then(function(component){
-            return component.rdfpf.createVni;
+        return Promise.resolve(createComponent()).then(function(component){
+            return component.rpf.createVni;
         }).should.be.a('function');
     });
     it("createVni should return an object", function() {
-        return Promise.resolve(component).then(function(component){
-            return component.rdfpf.createVni('');
+        return Promise.resolve(createComponent()).then(function(component){
+            return component.rpf.createVni('');
         }).should.be.an('object');
     });
     describe('#createVni', function() {
         it("should have getInputState function", function() {
-            return Promise.resolve(component).then(function(component){
-                return component.rdfpf.createVni('').getInputState;
+            return Promise.resolve(createComponent()).then(function(component){
+                return component.rpf.createVni('').getInputState;
             }).should.be.a('function');
         });
         it("should have getOutputState function", function() {
-                return Promise.resolve(component).then(function(component){
-                return component.rdfpf.createVni('').getOutputState;
+                return Promise.resolve(createComponent()).then(function(component){
+                return component.rpf.createVni('').getOutputState;
             }).should.be.a('function');
         });
         describe('#getInputState', function() {
             it("should initially return nothing", function() {
-                return Promise.resolve(component).then(function(component){
-                    return component.rdfpf.createVni('').getInputState('patient', '', '');
+                return Promise.resolve(createComponent()).then(function(component){
+                    return component.rpf.createVni('').getInputState('patient', '', '');
                 }).should.become(undefined);
             });
             it("should have patient input state after input", function() {
-                return Promise.resolve(component).then(function(component){
+                return Promise.resolve(createComponent()).then(function(component){
                     var patient = noflo.internalSocket.createSocket();
                     component.inPorts.patient.attach(patient);
                     patient.send({id: '001',  name: 'Alice', dob: '1979-01-23' });
                     patient.disconnect();
                     component.inPorts.patient.detach(patient);
-                    return component.rdfpf.createVni('').getInputState('patient', '', '');
+                    return component.rpf.createVni('').getInputState('patient', '', '');
                 }).then(_.keys).then(_.sortBy).should.become(_.sortBy(['data','lm', 'previousLms', 'selfPort']));
             });
             it("should have patient input state data after input", function() {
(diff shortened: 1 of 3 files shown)


test/merge-patient-lab-iips-mocha.js
```diff
@@ -0,0 +1,138 @@
...
+                    component.inPorts.patient.attach(patient);
+                    patient.send({id: '001',  name: 'Alice', dob: '1979-01-23' });
+                    patient.disconnect();
+                    component.inPorts.labwork.attach(labwork);
+                    labwork.send({id: '001',  glucose: '75',  date: '2012-02-01'});
+                    labwork.disconnect();
+                    component.inPorts.patient.detach(labwork);
+                    component.inPorts.patient.detach(patient);
+                    return component.rdfpf.createVni('').getOutputState();
+                }).then(_.keys).then(_.sortBy).should.become(_.sortBy(['data','lm', 'previousLms', 'selfPort']));
```
dbooth-boston, line 115: Since the updater may be asynchronous, will this be a race condition?  Can we be sure that the updater would have run?

test/merge-patient-lab-iips-mocha.js
```diff
@@ -0,0 +1,138 @@
...
+                }).then(_.property('data')).should.become({id: '001',  name: 'Alice', dob: '1979-01-23', glucose: '75',  date: '2012-02-01'});
+            });
+        });
+    });
+});
+
+function createComponent(getComponent) {
+    var component = getComponent();
+    _.forEach(component.inPorts, function(port, name) {
+        port.nodeInstance = component;
```
dbooth-boston, line 141: I was going to suggest that we use the term nodeInstance instead of 'component' throughout, to match what NoFlo calls it.  But now that I think more about it, I prefer the name "component", so I suggest we use that term throughout (which is what this code is already doing).

test/merge-patient-lab-iips-mocha.js
```diff
@@ -0,0 +1,138 @@
...
+        port.name = name;
+    });
+    _.forEach(component.outPorts, function(port, name) {
+        port.nodeInstance = component;
```
dbooth-boston, line 145: Wouldn't the test be more realistic if a NoFlo server were run, instead of manually imitating what NoFlo does?  Or would that make testing too difficult?

test/merge-patient-lab-iips-mocha.js
```diff
@@ -0,0 +1,138 @@
...
+        }).should.be.an('object');
+    });
+    it("should have createVni function", function() {
+        return Promise.resolve(component).then(function(component){
+            return component.rdfpf.createVni;
+        }).should.be.a('function');
+    });
+    it("createVni should return an object", function() {
+        return Promise.resolve(component).then(function(component){
+            return component.rdfpf.createVni('');
```
dbooth-boston: createVni needs one more parameter (for split/join later): parentVni.  It can be undefined here.

test/merge-patient-lab-iips-mocha.js
```diff
@@ -0,0 +1,138 @@
...
+    var component = createComponent(componentFactory);
+    it("should have rdfpf object on nodeInstance", function() {
+        return Promise.resolve(component).then(function(component){
+            return component.rdfpf;
+        }).should.be.an('object');
+    });
+    it("should have createVni function", function() {
+        return Promise.resolve(component).then(function(component){
+            return component.rdfpf.createVni;
+        }).should.be.a('function');
```
dbooth-boston, line 21: I forgot to discuss on today's call: should we make these functions be properties of the various runtime objects, or just functions that are exported from the appropriate module?  For example, call component.rpf.createVni( ...) or call createVni(component, ...)?  If they are properties of the objects then the object uses more memory and requires property chain lookup.  On the other hand, I guess the benefit of being attached to the object is that the functions are bundled with the object.  http://www.toptal.com/javascript/javascript-prototypes-scopes-and-performance-what-you-need-to-know  But I would think that we could still bundle the functions into a module, using module.exports even if they were not attached to the runtime objects.  The memory usage issue would be more applicable to VNI and State objects than anything else.

It sort of seems to be that it would be better to NOT store the functions in the objects, but instead pass the object explicitly as the first parameter, like createVni(component, ...), avoiding the use of 'this'.  I do not want to do premature optimization here, but I do want to be aware of the trade-offs that we are making, and make the best trade-off.  I also would prefer to do things consistently one way or the other -- for sanity's sake -- and for that the explicit parameter passing style seems more general.  I have also found that it is considerably easier to read code that uses an explicit parameter instead of passing an object implicitly using 'this', because it is not easy to know what 'this' is supposed to be bound to.  In contrast, explicit parameters typically have descriptive names.

Some discussion here:
https://news.ycombinator.com/item?id=4985853

What do you think, and why?

test/merge-patient-lab-iips-mocha.js
glenna-m, line 115: Nope - updater may not have run.   I believe this will be an issue with this test.

test/merge-patient-lab-iips-mocha.js
glenna-m, line 141: This is a unit test.  We create a component so we reference it as the component.  In other situations, this may or may not always be the case.  Let's look at that as we get there.

test/merge-patient-lab-iips-mocha.js
glenna-m, line 145: This is a behavioral unit test, not a functional test yet.  The goal of behavioral unit testing is to isolate the test to just a particular behavior under test.    Functional testing comes later.

test/merge-patient-lab-iips-mocha.js
glenna-m, line 21: Let's discuss the next time we meet.

test/merge-patient-lab-iips-mocha.js
catch-point: I don't understand what parentVni would be.

glenna-m: It's for the Split case. Parent VNI is the vni of the node instance/component.  There would be a vni for each of the child subnodes, and those vnis would have the reference to the parent Vni of the node instance/component that contains them.

catch-point: Why is that needed? Since the parent vni will always have an vnid of '', what value would there be by passing it in here?

glenna-m: It's not needed for this Sprint.  David wants it for a future sprint where we have non-IIP input and start handling split.  It came up because we are working with the API that will need it.

test/merge-patient-lab-iips-mocha.js
dbooth-boston, line 145: ok

test/merge-patient-lab-iips-mocha.js
catch-point, line 115: I have added a delay that should address this

PR #17, opened 2016-02-02
Add change-component-factory

catch-point:
Trigger change event with matching payloads (including output payloads) when in port payloads change

### DIFF
--- a/components/change-component-factory.js
+++ b/components/change-component-factory.js
@@ -1,0 +1,126 @@
+// onchange-component-factory.js
+
+var _ = require('underscore');
+var jsonpointer = require('jsonpointer');
+var componentFactory = require('./event-component-factory');
+
+/**
+ * Combines payloads, from multiple ports, arriving in an asynchronous
+ * fashion. Each port may include an indexBy function, property, or path
+ * that will be used to determine a key that must match payloads from other ports.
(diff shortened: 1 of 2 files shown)


dbooth-boston:
I'm uncomfortable with this approach, for a few reasons: 1. I prefer that we expect and use the payload.vnid property directly, rather than using a general indexBy mechanism, because it would simplify the code, and I don't think we need the generality of indexBy.  2. I think the payload.state should be saved into the inputStates object directly, instead saving payloads in one step, and then  saving them into the inputStates object in a later step.  3. It seems odd to me that this code also defines a component factory, since event-component-factory.js already did that.  I don't see why this module needs to talk about creating components if its purpose is to process events.

Let's discuss this on Thursday.  To my mind there should be only one onData function for the Framework, as outlined in https://docs.google.com/document/d/1mb_8DCV7nnvhYwc0Ns_gCWpzyXcM1_8qs2u9GJpIZP4/edit# , and it should implement the main algorithm described there, calling various helper functions to delegate portions of the work.  If the code structure is radically different than that, then it becomes very hard for me to see how it corresponds to that design.

glenna-m:
I believe this one is on hold for now, pending our new work integration, yes?

catch-point:
yes

PR #18, opened 2016-02-02
Add output-component-factory

catch-point:
Defines the common output and error ports and redirects thrown errors to the error port and result or array of results to the output port.

### DIFF
--- a/components/output-component-factory.js
+++ b/components/output-component-factory.js
@@ -1,0 +1,74 @@
+// output-component-factory.js
+
+var _ = require('underscore');
+var componentFactory = require('./change-component-factory');
+
+/**
+ * Captures thrown errors from the onchange handler and sends them to the
+ * 'error' port, and sends the result (maybe a Promise) of the handler function to
+ * the 'output' port.
+ *
+ * Usage:
+ *  componentFactory({
+ *      inPorts: {input:{indexBy:'vnid'}} || ['input'],
+ *      onchange: function({portName:payload})
+ *  });
+ */
+module.exports = function(nodeDef){
+    return componentFactory(_.defaults({
+        inPorts: _.isArray(nodeDef.inPorts) ?
+            _.object(nodeDef.inPorts, _.times(nodeDef.inPorts.length, _.object)) :
+            nodeDef.inPorts,
+        outPorts: _.extend({
+            output: {
+                description: "Resolved result of the updater",
+                datatype: 'object'
+            },
+            error: {
+                description: "Rejected error of the updater",
+                datatype: 'object'
+            }
+        }, nodeDef.outPorts),
+        onchange: _.wrap(nodeDef.onchange, wrapChangeHandler)
+    }, nodeDef));
+};
+
+/**
+ * Wrap the onchange handler function in a Promise to capture errors and return objects.
+ * @this Component
+ * @param handler onchange function
+ * @param payloads hash of portNames to payloads
+ */
+function wrapChangeHandler(handler, payloads) {
+    return Promise.resolve(payloads).then(handler.bind(this))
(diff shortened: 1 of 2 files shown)


components/output-component-factory.js
```diff
@@ -0,0 +1,74 @@
...
+ *      inPorts: {input:{indexBy:'vnid'}} || ['input'],
+ *      onchange: function({portName:payload})
+ *  });
+ */
+module.exports = function(nodeDef){
+    return componentFactory(_.defaults({
+        inPorts: _.isArray(nodeDef.inPorts) ?
+            _.object(nodeDef.inPorts, _.times(nodeDef.inPorts.length, _.object)) :
+            nodeDef.inPorts,
+        outPorts: _.extend({
```
glenna-m, line 22: Should this be _.defaults?   The updater developer could choose to define their own.

test/output-component-factory-mocha.js
```diff
@@ -0,0 +1,111 @@
...
+var chaiAsPromised = require('chai-as-promised');
+chai.should();
+chai.use(chaiAsPromised);
+
+var _ = require('underscore');
+var noflo = require('noflo');
+var componentFactory = require('../components/output-component-factory');
+
+describe('output-component-factory', function() {
+    it("should accept array of inPort names", function() {
```
glenna-m, line 13: Should have a test for when there are no inPorts specified.

test/output-component-factory-mocha.js
```diff
@@ -0,0 +1,111 @@
...
+            inPorts:['input'],
+            onchange:function(){}
+        }).then(componentFactory).then(createComponent).then(function(component){
+            return _.keys(component.inPorts);
+        }).should.eventually.contain('input');
+    });
+    it("should resolve onchange function", function() {
+        return Promise.resolve({
+            inPorts:['input'],
+            onchange: function(payloads) {
```
glenna-m, line 24: Test case for when there is no onchange function specified?

components/output-component-factory.js
```diff
@@ -0,0 +1,74 @@
...
+}
+
+/**
+ * Handle result of onchange function call.
+ * @this Component with this.outPorts.output Port
+ */
+function sendOutput(resolved) {
+    var port = this.outPorts.output;
+    if (_.isArray(resolved)) {
+        port.connect();
```
glenna-m, line 54: Not entirely sure what you are doing in this block  It looks like if you got a resolved parameter that is an array, you are connecting to the port, and binding the port on to each send call, then disconnecting.  I don't actually see you send anything.  Can you add a comment why you are doing this and who would be using the send function with that bound port?

test/output-component-factory-mocha.js
```diff
@@ -0,0 +1,111 @@
...
+                throw payloads.input + " world";
+            }
+        }).then(componentFactory).then(createComponent).then(function(component){
+            var console_error = console.error;
+            return new Promise(function(resolve, reject) {
+                console.error = reject;
+                var output = noflo.internalSocket.createSocket();
```
glenna-m, line 68: Can we make this block (send data on output port once) into a common helper function since it is used a fair bit?

test/output-component-factory-mocha.js
```diff
@@ -0,0 +1,111 @@
...
+        });
+        _.forEach(component.outPorts, function(port, name) {
+            port.nodeInstance = component;
+            port.name = name;
+        });
+        return component;
+    }
+    function sendData(component, port, payload) {
+        var socket = noflo.internalSocket.createSocket();
+        component.inPorts[port].attach(socket);
```
glenna-m, line 95: Could you use common-test.sendData here?

PR #20, opened 2016-02-03
Glenna sprint5

glenna-m:


components/javascript-wrapper.js
```diff
@@ -0,0 +1,157 @@
+/**
+ * This file contacts the common code used by all of the variations of any RDF
```
dbooth-boston: Is this comment correct?

components/javascript-wrapper.js
```diff
@@ -0,0 +1,157 @@
...
+
+/**
+ * This module provides a Javascript wrapper for updaters.  It is responsible for setting up the
+ * code to ensure updater callback, and then invoking the RDF pipeline framework to complete building
+ * the component and setting up the vni and state metadata.
+ *
+ * @param nodeDef updater node definition. This can be a simple updater function, or an object that contains 
+ *        updater metadata (e.g., input/output port definitions, description, icons), and an updater.
+ *
+ * @return a promise to create the noflo rdf component
```
dbooth-boston, line 18: Why not create the noflo component directly, instead of returning a promise to do it?  I can see the need for Promises for asynchronous code that is executed at runtime (i.e., when the pipeline is running).  But he process of creating the components for the pipeline can be entirely synchronous, can't it?  Why are Promises needed for that?

components/javascript-wrapper.js
```diff
@@ -0,0 +1,157 @@
...
+ */
+module.exports = function(nodeDef) { 
+
+    // Save the caller's node instance 
+    var self = this;
```
dbooth-boston, line 23: What is 'this' expected to be bound to, when this is run?  In general, please document all uses of 'this' with an @this comment above the function: https://en.wikipedia.org/wiki/JSDoc#JSDoc_tags

components/javascript-wrapper.js
```diff
@@ -0,0 +1,157 @@
...
+ * the component and setting up the vni and state metadata.
+ *
+ * @param nodeDef updater node definition. This can be a simple updater function, or an object that contains 
+ *        updater metadata (e.g., input/output port definitions, description, icons), and an updater.
+ *
+ * @return a promise to create the noflo rdf component
+ */
+module.exports = function(nodeDef) { 
+
+    // Save the caller's node instance
```
dbooth-boston, line 22: Is this comment correct?  I would think that the nodeDef parameter would be the definition of the node that we wish to create, so I am confused about what node instance this would be talking about.  I am confused because if nodeDef is being passed in, then presumably the node instance has not yet been created.

components/javascript-wrapper.js
```diff
@@ -0,0 +1,157 @@
...
+
+    // Save the caller's node instance 
+    var self = this;
+    
+    // Ensure we have at least one input port and and updater, using a default stub updater if 
+    // there is none specified
+    var updaterArgs = null;
+    if ( _.isUndefined(nodeDef) || 
+       ( ! _.isFunction(nodeDef) && ( _.isEmpty( nodeDef ) || ( _.isEmpty( nodeDef.inPorts ))))) {
+       // Got nothing to work with - no input ports and no updater, so create a default updater and input port
```
dbooth-boston, line 30: Excellent

components/javascript-wrapper.js
```diff
@@ -0,0 +1,157 @@
...
+       ( ! _.isFunction(nodeDef) && ( _.isEmpty( nodeDef ) || ( _.isEmpty( nodeDef.inPorts ))))) {
+       // Got nothing to work with - no input ports and no updater, so create a default updater and input port 
+       nodeDef = { 
+           inPorts: {
+               input: {
+                   datatype: 'object',
+                   description: "default input port" 
+               }
+           },
+           updater: defaultUpdater
```
dbooth-boston, line 38: Suggestion: rename defaultUpdater -> defaultJavascriptUpdater, because there will be other kinds of updaters, and this one is for the JavascriptWrapper.

components/javascript-wrapper.js
```diff
@@ -0,0 +1,157 @@
...
+       };
+
+    }  else if ( _.isUndefined( nodeDef.updater ) && !_.isFunction(nodeDef)) { 
+       // Got some node metadata, but no node updater - use the metadata with default updater
```
dbooth-boston, line 42: good

components/javascript-wrapper.js
```diff
@@ -0,0 +1,157 @@
...
+       nodeDef = _.extend( nodeDef, 
+                           { updater: defaultUpdater } );
+
+    } else if ( _.isFunction( nodeDef ) ) { 
+        // Got an updater but no metadata - use updater args as input port list
+        // This assumes that all updater arguments are input
```
dbooth-boston, line 48: Good.  One change though, please: allow output as a parameter name.  If 'output' is one of the parameters, then pass vni.output.data to it when the updater is called, so that the updater can easily access the node's current output state data via that parameter.

components/javascript-wrapper.js
```diff
@@ -0,0 +1,157 @@
...
+        updaterArgs = introspect( nodeDef );
+        nodeDef = { inPorts: updaterArgs,
+                    updater: nodeDef };
+    } 
+
+    // Have an updater now, no matter what, be it default or not - figure out what the updater arguments are
+    var updaterArgs = (_.isNull( updaterArgs )) ? introspect( nodeDef.updater ) : updaterArgs;
```
dbooth-boston, line 55: The introspect function was already called above.  Can't it be called in only one place instead of two?  Maybe only here instead of line51?

components/javascript-wrapper.js
glenna-m, line 18: I knew that James was planning to overhaul the lower layers and I wanted to focus on the vnis, state, and javascript wrapper for this sprint -> using this was the fastest way to get to work on what I wanted to work on.  My plan is to replace this call with calls to James' new code.

components/javascript-wrapper.js
glenna-m, line 23: Will do.  For the record, when run from noflo-server, "this" is the node.js instance.

components/javascript-wrapper.js
glenna-m, line 22: The comment is correct - this is not new.

components/javascript-wrapper.js
```diff
@@ -0,0 +1,157 @@
...
+                           }, nodeDef)
+                     );
+
+}; // module.exports
+
+// Define a default updater stub function to be used if the caller 
+// did not pass one in.  Simply returns whatever the original input was so it
+// can be sent through the output port to the next component
+var defaultUpdater = function() { 
+    return { out: arguments };
```
dbooth-boston: Excellent.  However, since we changed to allow only one state per node, the updater API can be simplified to just return the new output state data, instead of including the port name.  For example, if there is only one updater parameter 'input', then an updater could be written:

<pre>
function myUpdater(input) { return input; }
</pre>


We didn't discuss what inputs the defaultJavascriptUpdater should have, or what it should do. How about if we do the following:

 <pre>
1. Automatically add an inPort for every updater parameter 
(except 'output') that is not already listed in inPorts.
2. if (no updater and no inPorts) { add port 'input' to inPorts }
3. if (no updater) { 
     Make a default updater, using the inPort names as parameters.
     If there is only one inPort, have it return that data directly.
     Otherwise, have it return an object whose keys are the parameter
     (inport) names, and whose values are the values of those 
     parameters.  For example, if the updater has two parameters in1
     and in2, then it would return: { in1: in1, in2: in2 }
}
</pre>

components/javascript-wrapper.js
```diff
@@ -0,0 +1,157 @@
...
+
+        return args;
+    } catch (e) { 
+        // Could throw exception if we do not have source to introspect
+        // Use default of empty argument list
+        return [];
+    }
+};
+
+var ondata = function( portName, updaterArgs, nodeDef, payload ) {
```
dbooth-boston: Is ondata only being defined temporarily here, because the underlying Framework layer is not written yet?  Once the Framework layer has been implemented, the ondata function should only be defined there -- not in any wrapper.
The only API between the wrapper and the Framework should be through the three functions fExists, fRunUpdater and fDestroyState: https://github.com/rdf-pipeline/noflo-rdf-pipeline/wiki/Wrapper-API

components/javascript-wrapper.js
glenna-m, line 38: Do you expect multiple updaters for a node?   If not, defaultUpdater is probably a better choice since it can be used for any type.    Architecturally, you know what wrapper it is, since your component developer specified it.   Having a generic defaultUpdater is more flexible and consistent, no?  Let's talk about this one.

components/javascript-wrapper.js
```diff
@@ -0,0 +1,157 @@
...
+           // No updaterArgs, so just pass the state 
+           return nodeDef.updater.call( this, this.rpf.state );
+
+        } else {
+
+           // Have one or more parameters - get them in the right order and pass them 
+           var parameters = _.map( updaterArgs, function( arg ) { 
+               var parameter = vni.inputState( arg, 'test', '*' );
+               if ( _.isUndefined(parameter) && nodeDef.outPorts && nodeDef.outPorts.hasOwnProperty( arg ) ) {
+                   // passing in the name of the output port - is this correct?
```
dbooth-boston: If one of the parameters is called 'output', then vni.output.data should be passed as its value.

components/javascript-wrapper.js
glenna-m, line 48: Yes, I agree.  But I knew the promise component was expecting an out port and I wanted to see what James was doing before making changes in that area.

components/javascript-wrapper.js
dbooth-boston, line 18: ok

components/javascript-wrapper.js
dbooth-boston, line 22: ok.  I guess i'll need this explained to me sometime.

components/javascript-wrapper.js
dbooth-boston, line 38: Only one updater per node.  Different wrappers will have different updater APIs.  The default updater for one wrapper will typically look different than the default wrapper for another wrapper.  That's why I was suggesting being more specific in the name.

components/javascript-wrapper.js
glenna-m: Nope.  Deleted it for next check in.  It was a hangover from an earlier version.

components/javascript-wrapper.js
glenna-m, line 22: Take a look at where it's called from in the components/merge-patient-lab-iips.js. Whatever the updater writer defines as the node definition will get passed in to the wrapper.

components/javascript-wrapper.js
glenna-m, line 38: The idea was to put it in a standard place so it's easy to pick up any wrapper and know where the updater is defined what ever it is.  This would simplify support.   You can put whatever you want in there - it's up to the wrapper to interpret it.

components/javascript-wrapper.js
glenna-m, line 55: Introspect is not being called twice.  There are 3 paths through the if/else above, and only one will call the introspect function.  If it has been called, updaterArgs will not be null, so we'll just use whatever is in there (the updaterArgs reference after the colon).    If it was not called, only then, would we call it.

components/javascript-wrapper.js
glenna-m: That code is already doing #1 and #2, no?  

For the return, let's talk - that's inconsistent with the regular path.  Wouldn't it be better to return the something more like {out: { in1: in1, in2: in2}} ?  That would allow forwarding through the out port work work. (naturally, output port when we roll to that).

components/javascript-wrapper.js
glenna-m: Yes, I know the API.

The goal of this code was to meet Sprint objectives and not overlap James's lower level changes.     The only hook i had in the promiseComponent was ondata.

components/javascript-wrapper.js
glenna-m: I was not trying to roll to the new output definition in this version.

components/javascript-wrapper.js
dbooth-boston, line 55: ok, thanks

components/javascript-wrapper.js
dbooth-boston: It looks like line 51 does #1 if no inPorts were specified, but l think it would be good to do it even if some inPorts were specified.  In other words, for any updater parameter (other than 'output') that is not already listed in inPorts, add it to inPorts.

For #2, yes I now see that line 35 does that already.

For the return, the question is how we want users to be able to write their updaters.  I'm thinking that it would be simpler for users (since a node now can only have one output state) if the updater could just write:

<pre>
return newData;
</pre>

instead of

<pre>
return {output: newData};
</pre>

but we can discuss this.  The downside of returning newData directly is that we still need some way for the updater to indicate that the state did not change, such as:

<pre>
return rpf.NO_CHANGE;
</pre>

That doesn't look very pretty, but it is explicit.  See slide 6, which I just updated:
http://dbooth.org/2013/pipeline/RDF-Pipeline-Implementation.pdf

components/javascript-wrapper.js
dbooth-boston: Ok, thanks.

components/javascript-wrapper.js
dbooth-boston: ok

components/state.js
```diff
@@ -0,0 +1,86 @@
+// state.js
+
+var _ = require('underscore');
+
+var Lm = require('./Lm');
+
+var inputStates = {}; 
+var outputStates = {};
```
dbooth-boston: outputStates will be replaced by vni.output and vni.error, though I don't know if you intend to address that in this sprint.

components/state.js
```diff
@@ -0,0 +1,86 @@
...
+     * return a hash of all input states for every port and its sources 
+     */
+    allOutputStates: function() { 
+        return outputStates;
+    },
+
+    /**
+     * Get/set the current state for the specified input port from the originating port
+     */
+    inputState: function( inputPortName, originatingNodeId, originatingPortName, payload ) {
```
dbooth-boston: To avoid confusion with NoFlo payloads, I suggest renaming payload -> newState, where newState is a State object (with lm and data properties).

Also, how about renaming originatingNodeId -> originatingNodeName (for consistency with inputPortName and originatingPortName)?

components/state.js
```diff
@@ -0,0 +1,86 @@
...
+
+        if ( inputPortName && originatingNodeId ) {
+
+            originatingPortName = 
+                (_.isUndefined( originatingPortName) || _.isEmpty( originatingPortName )) ? IIP_PORT : originatingPortName;
+
+            if ( _.isUndefined( inputStates[inputPortName] ) ) { 
+                inputStates[inputPortName] = {};
```
dbooth-boston: good

components/state.js
```diff
@@ -0,0 +1,86 @@
...
+            }
+            if ( _.isUndefined( inputStates[inputPortName][originatingNodeId] ) ) {
+                inputStates[inputPortName][originatingNodeId] = {};
```
dbooth-boston: good

components/state.js
```diff
@@ -0,0 +1,86 @@
...
+
+    /**
+     * Get/set the current state for the specified input port from the originating port
+     */
+    inputState: function( inputPortName, originatingNodeId, originatingPortName, payload ) {
+
+        if ( inputPortName && originatingNodeId ) {
+
+            originatingPortName = 
+                (_.isUndefined( originatingPortName) || _.isEmpty( originatingPortName )) ? IIP_PORT : originatingPortName;
```
dbooth-boston: Shouldn't we have the same kind of statement for originatingNodeId, to give it a defined value in the case of an IIP?  Otherwise, if it is undefined, it will become the string 'undefined' when used as a hash key.
https://es5.github.io/#x9.8
Actually, I guess the code above would work.  But it seems odd to ensure that originatingPortName is set, but not originatingNodeName.  For simplicity and uniformity, how about just saying:
<pre>
originatingPortName = originatingPortName || '';
originatingNodeName = originatingNodeName || '';
<pre>

components/state.js
```diff
@@ -0,0 +1,86 @@
...
+    allOutputStates: function() { 
+        return outputStates;
+    },
+
+    /**
+     * Get/set the current state for the specified input port from the originating port
+     */
+    inputState: function( inputPortName, originatingNodeId, originatingPortName, payload ) {
+
+        if ( inputPortName && originatingNodeId ) {
```
dbooth-boston: I don't see why this 'if' statement is needed.  

Also, if inputPortName is undefined then an error should be thrown, because the function was called wrong.

components/state.js
```diff
@@ -0,0 +1,86 @@
...
+
+        } else { 
+            return;
+        }
+    },
+ 
+    /**
+     * Get/set the current state for the specified output port
+     */
+    outputState: function( outPortName, state ) {
```
dbooth-boston: This will be dropped in a later commit, when the VNI data structure is changed to have vni.output and vni.error, right?

components/vni.js
```diff
@@ -0,0 +1,63 @@
...
+var Lm = require('./Lm');
+var state = require('./state');
+
+// Hash of vnis
+vnis = []; 
+
+// Define Vni class and its subclasses - InterNodeVni & IipVni
+function Vni( vnid ) {
+  this.vnid = vnid;
+  this.lm = Lm();
```
dbooth-boston, line 13: A VNI does not have an lm.  A State has an lm, but should be left undefined initially.

components/state.js
glenna-m: I was requiring an originating Node ID so it would not be undefined.    This line (#44) ensured that: 
     if ( inputPortName && originatingNodeId ) {
The lines above won't do what this code is doing.   The proposed code for the port would allow an empty originatingPortName to be used as a hash key, which is illegal.   I'm forcing either an undefined or empty originating port to be '*' here so it doesn't cause a failure when we try to use it in the hash below.

components/state.js
glenna-m: Will make both of these changes.

components/vni.js
```diff
@@ -0,0 +1,63 @@
...
+  this.allInputStates = state.allInputStates;
```
dbooth-boston: I am uncomfortable having member functions stored in VNIs and States, because they consume memory and there may be many VNI and State objects when scaling up.  For encapsulation, functions like allOutputStates can be in this VNI module (kind of like static methods), but I don't see any real need to store them in the VNI objects themselves.  I would rather keep the VNI and State objects as lean as possible.

components/state.js
glenna-m: I'll remove the outputStates from here.   Thanks.

components/vni.js
```diff
@@ -0,0 +1,63 @@
...
+  this.allOutputStates = state.allOutputStates;
+  this.inputState = state.inputState;
+  this.outputState = state.outputState;
+  this.node = this.nodeInstance;
+}
+
+function NodePortVni( vnid ) {
```
dbooth-boston, line 21: What is the purpose of this function?

components/vni.js
```diff
@@ -0,0 +1,63 @@
...
+  Vni.call(this, vnid);
+  this.previousLms = [];
```
dbooth-boston, line 23: Is 'this' a State object?  I assumed it would be a VNI.  previousLms is a property of a State object, not of a VNI.

components/vni.js
```diff
@@ -0,0 +1,63 @@
...
+}
+
+var IIP_VNID = '*';
```
dbooth-boston, line 26: The default vnid must be the empty string, because any other string could be a legitimate vnid created by the user.

components/vni.js
```diff
@@ -0,0 +1,63 @@
...
+function IipVni() {
```
dbooth-boston, line 27: I don't think we need to create a VNI for an IIP, but we  do need to create a State object for an IIP.  If the IIP data is someData, then the State and Payload would be:

<pre>
if (!payload.vnid) {
  // Received data from an IIP.  Convert the payload to a regular payload.
  var state = { lm: Lm(), data: payload };
  payload = { vnid: '', state: state };
}

// Eventually gets stored as:
vnis[vnid].inputStates[inPortname][originatingNodeName][originatingPortName] = payload.state;
</pre>

components/vni.js
glenna-m, line 13: will remove the lm.

components/change-component-factory.js
```diff
@@ -0,0 +1,130 @@
...
+ *
+ * When payload is received on an indexed port change will be triggered with both
+ * indexed payload and non-indexed payload. However, when payload is received on
+ * a non-indexed port, change event is only triggered with other non-indexed payloads.
+ *
+ * Usage:
+ *   componentFactory({
+ *     inPorts: {input:{indexBy:'vnid'}},
+ *     outPorts: {output:{indexBy:'vnid'}},
+ *     onchange: function({input:payload,output:payload})
```
dbooth-boston, line 23: We should discuss this.

components/javascript-wrapper.js
dbooth-boston, line 22: ok.  clarified.  thanks!

components/javascript-wrapper.js
dbooth-boston, line 23: thanks!

components/javascript-wrapper.js
dbooth-boston, line 38: ok.  fine to leave it as defaultUpdater.

components/vni.js
```diff
@@ -0,0 +1,63 @@
...
+
+// Define Vni class and its subclasses - InterNodeVni & IipVni
+function Vni( vnid ) {
+  this.vnid = vnid;
+  this.lm = Lm();
+  this.allInputStates = inputs.allInputStates;
+  this.inputState = inputs.inputState;
+  this.output = {};
+  this.error = {};
+  this.node = this.nodeInstance;
```
dbooth-boston, line 18: how about adding a comment:

// this.parentVni = undefined;

components/vni.js
```diff
@@ -0,0 +1,63 @@
...
+var inputs = require('./input-state');
+
+// Hash of vnis
+vnis = []; 
+
+// Define Vni class and its subclasses - InterNodeVni & IipVni
+function Vni( vnid ) {
+  this.vnid = vnid;
+  this.lm = Lm();
+  this.allInputStates = inputs.allInputStates;
```
dbooth-boston, line 14: Move the functions out of the object, so that the VNI and State objects are as lean as possible -- only hashes of data properties.

components/vni.js
```diff
@@ -0,0 +1,63 @@
...
+  this.inputState = inputs.inputState;
```
dbooth-boston, line 15: Add: this.inputStates = createInputStates(...)

components/vni.js
dbooth-boston, line 26: ACTION: David to test empty string key, and three-level hash

components/javascript-wrapper.js
catch-point, line 23: Given that 'self' is never used, I think this line should be removed as it is a little confusing to have it.

components/javascript-wrapper.js
```diff
@@ -0,0 +1,131 @@
...
+    }  else if ( _.isUndefined( nodeDef.updater ) && !_.isFunction(nodeDef)) { 
+       // Got some node metadata, but no node updater - use the metadata with default updater
+       nodeDef = _.extend( nodeDef, 
+                           { updater: defaultUpdater } );
+
+    } else if ( _.isFunction( nodeDef ) ) { 
+        // Got an updater but no metadata - use updater args as input port list
+        // This assumes that all updater arguments are input
+        updaterArgs = introspect( nodeDef );
+        nodeDef = { inPorts: updaterArgs,
```
catch-point, line 50: You should exclude 'output' and 'error' from the set of possible names in inPorts.

components/javascript-wrapper.js
```diff
@@ -0,0 +1,131 @@
...
+                    updater: nodeDef };
+    } 
+
+    // Have an updater now, no matter what, be it default or not - figure out what the updater arguments are
+    var updaterArgs = (_.isNull( updaterArgs )) ? introspect( nodeDef.updater ) : updaterArgs; 
+
+    return componentFactory( 
+        _.defaults(
+             { onchange: function(args) {
```
catch-point, line 59: You can use _.partial(onchange, updaterArgs) here instead, to avoid an inline function.

components/javascript-wrapper.js
```diff
@@ -0,0 +1,131 @@
...
+             }, 
+             nodeDef)
+        );
+}; // module.exports
+
+// Define a default updater stub function to be used if the caller 
+// did not pass one in.  Simply returns whatever the original input was so it
+// can be sent through the output port to the next component
+var defaultUpdater = function() { 
+    return { output: arguments };
```
catch-point, line 72: My understanding is that the output port will be assumed and not included in the result.

components/promise-component.js
```diff
@@ -62,13 +62,15 @@ module.exports = function(def){
             }
         }, port);
     });
+    
     return function() {
         return _.extend(new noflo.Component({
-            outPorts: _.defaults(_.object(
```
catch-point, line 67: Why is this included in this PR?

test/merge-patient-lab-iips-mocha.js
```diff
@@ -0,0 +1,147 @@
...
+var chaiAsPromised = require('chai-as-promised');
+chai.should();
+chai.use(chaiAsPromised);
+
+var _ = require('underscore');
+
+var noflo = require('noflo');
+var sinon = require('sinon');
+
+var componentFactory = require('../components/merge-patient-lab-iips');
```
catch-point, line 13: Changes to the sprint test case should be in a separate PR. This way it is clear that this PR adheres to the agreed upon behaviour.

components/change-component-factory.js
```diff
@@ -0,0 +1,130 @@
...
+ */
+function indexPayload(indexBy, portName, payload) {
+    // determine a key string or undefined if no key should be used
+    var key = indexBy == null ? undefined : ('' + indexBy.call(this, payload));
+    var rpf = this.nodeInstance.rpf = this.nodeInstance.rpf || {};
+    var payloads = rpf.payloads || {};
+    var index = key == null ? payload :
+        _.extend(payloads[portName] || {}, _.object([key], [payload]));
+    rpf.payloads = _.extend(payloads, _.object([portName], [index]));
+    var vni = rpf.vni('').inputState( portName, 'test', '*', rpf.payloads );
```
catch-point, line 90: Changes to this file should be make in the change-component-factory PR, not this one.

components/event-component-factory.js
```diff
@@ -0,0 +1,115 @@
...
+            inPorts: _.mapObject(def.inPorts, _.clone)
+        });
+        triggerPortDataEvents(nodeInstance.outPorts);
+        registerPorts(nodeInstance.outPorts, def.outPorts);
+        registerPorts(nodeInstance.inPorts, def.inPorts);
+        registerListeners(nodeInstance, def);
+        nodeInstance.description = def.description;
+        nodeInstance.setIcon(def.icon);
+        nodeInstance.rpf = rpf;
+        if ( ! _.isUndefined( def.updater ) ) {
```
catch-point, line 44: Changes to this file should be made in the event-component-factory PR

components/vni.js
catch-point, line 18: I prefer to only include code that is needed to complete this sprint and not sprinkle code with features that have not been fully discussed/fleshed out.

components/vni.js
```diff
@@ -0,0 +1,63 @@
+// Vni.js
+
+var _ = require('underscore');
+var Lm = require('./Lm');
+var inputs = require('./input-state');
+
+// Hash of vnis
+vnis = [];
```
catch-point, line 8: What is the scope of vnis? I think it should be specific to the nodeInstance and not be a global variable.

components/vni.js
dbooth-boston, line 18: ok with me

components/vni.js
dbooth-boston, line 26: Followup: It works.  But you have to remember to explicitly create the intermediate hashes, unlike in Perl, which will create them automagically.

PR #21, opened 2016-02-05
Behavioral tests with some basic stubs to enable testing.

glenna-m:
Behavioural test changes with stubs

### DIFF
--- a/components/inputStates.js
+++ b/components/inputStates.js
@@ -1,24 +1,21 @@
-// inportStates.js
+// inputStates.js
 
 var _ = require('underscore');
 var Lm = require('./Lm');
 
-var inportStates = [];
+var inputStates = [];
 
 // Returns the vni associated with the specified vnid.
 module.exports = function( inportName, sourceNodeName, sourcePortName, newState ) { 
  
    if ( arguments.length === 1 || arguments.length === 3 ) { 
-       if ( _.isUndefined( inportStates[inportName] ) ) { 
-           return;
-       } 
-       return inportStates[inportName];
+       return inputStates[inportName];
    }
 
    if ( arguments.length === 2 || arguments.length === 4 ) { 
-       inportStates[inportName] = { data: arguments[arguments.length-1],
-                                    lm: Lm(),
-                                    port: null  };
+       inputStates[inportName] = { data: arguments[arguments.length-1].data,
+                                   lm: arguments[arguments.length-1].lm };
+       // Will handle sourceNodeName and sourcePortName here in a future sprint
        return this;
    }
 };
--- a/components/inputStates.js
+++ b/components/inputStates.js
@@ -1,0 +1,24 @@
+// inportStates.js
+
+var _ = require('underscore');
+var Lm = require('./Lm');
+
+var inportStates = [];
+
+// Returns the vni associated with the specified vnid.
+module.exports = function( inportName, sourceNodeName, sourcePortName, newState ) { 
+ 
+   if ( arguments.length === 1 || arguments.length === 3 ) { 
+       if ( _.isUndefined( inportStates[inportName] ) ) { 
+           return;
+       } 
+       return inportStates[inportName];
+   }
+
+   if ( arguments.length === 2 || arguments.length === 4 ) { 
+       inportStates[inportName] = { data: arguments[arguments.length-1],
+                                    lm: Lm(),
+                                    port: null  };
+       return this;
+   }
+};
--- a/components/inputStates.js
+++ b/components/inputStates.js
@@ -13,8 +13,7 @@ module.exports = function( inportName, sourceNodeName, sourcePortName, newState 
    }
 
    if ( arguments.length === 2 || arguments.length === 4 ) { 
-       inputStates[inportName] = { data: arguments[arguments.length-1].data,
-                                   lm: arguments[arguments.length-1].lm };
+       inputStates[inportName] = arguments[arguments.length-1];
        // Will handle sourceNodeName and sourcePortName here in a future sprint
        return this;
    }
--- a/components/outputStates.js
+++ b/components/outputStates.js
@@ -1,23 +1,20 @@
-// outportStates.js
+// outputStates.js
 
 var _ = require('underscore');
 var Lm = require('./Lm');
+var State = require('./state');
 
-var outportStates = [];
+var outputStates = [];
 
 // Returns the vni associated with the specified vnid.
-module.exports = function( outportName, newState ) { 
+module.exports = function( outportName, data ) { 
  
    if ( arguments.length === 1 ) { 
-       if ( _.isUndefined( outportStates[outportName] ) ) { 
-           return;
-       } 
-       return outportStates[outportName];
+       return outputStates[outportName];
    }
 
    if ( arguments.length === 2 ) { 
-       outportStates[outportName] = { lm: Lm(),
-                                      data: newState };
+       outputStates[outportName] = State( data );
        return this;
    }
 };
--- a/components/outputStates.js
+++ b/components/outputStates.js
@@ -7,14 +7,14 @@ var State = require('./state');
 var outputStates = [];
 
 // Returns the vni associated with the specified vnid.
-module.exports = function( outportName, data ) { 
+module.exports = function( outportName, newState ) { 
  
    if ( arguments.length === 1 ) { 
        return outputStates[outportName];
    }
 
    if ( arguments.length === 2 ) { 
-       outputStates[outportName] = State( data );
+       outputStates[outportName] = newState;
        return this;
    }
 };
--- a/components/outputStates.js
+++ b/components/outputStates.js
@@ -1,0 +1,23 @@
+// outportStates.js
+
+var _ = require('underscore');
+var Lm = require('./Lm');
+
+var outportStates = [];
+
+// Returns the vni associated with the specified vnid.
+module.exports = function( outportName, newState ) { 
+ 
+   if ( arguments.length === 1 ) { 
+       if ( _.isUndefined( outportStates[outportName] ) ) { 
+           return;
+       } 
+       return outportStates[outportName];
+   }
+
+   if ( arguments.length === 2 ) { 
+       outportStates[outportName] = { lm: Lm(),
+                                      data: newState };
+       return this;
+   }
+};
--- a/components/state.js
+++ b/components/state.js
@@ -1,0 +1,26 @@
+// state.js
+
+var _ = require('underscore');
+
+var Lm = require('./Lm');
+
+var states = [];
+
+// Returns the vni associated with the specified vnid.
+module.exports = function( inportName, sourceNodeName, sourcePortName, newState ) { 
+ 
+   if ( arguments.length === 1 || arguments.length === 3 ) { 
+       if ( _.isUndefined( states[inportName] ) ) { 
+           return;
+       } 
+       return states[inportName];
+   }
+
+   if ( arguments.length === 2 || arguments.length === 4 ) { 
+       states[inportName] = { data: arguments[arguments.length-1],
+                              lm: Lm(),
+                              previousLms: [],
+                              port: inportName };
+       return this;
+   }
+};
--- a/components/state.js
+++ b/components/state.js
@@ -3,8 +3,9 @@
 var Lm = require('./Lm');
 
 // Returns the vni associated with the specified vnid.
-module.exports = function( data ) { 
+module.exports = function( data, lm ) { 
 
+       var stateLm = lm || Lm();
        return { data: data, 
-                lm: Lm() };
+                lm: stateLm };
 };
--- a/components/state.js
+++ b/components/state.js
@@ -1,26 +1,10 @@
-// state.js
-
-var _ = require('underscore');
+// State.js
 
 var Lm = require('./Lm');
 
-var states = [];
-
 // Returns the vni associated with the specified vnid.
-module.exports = function( inportName, sourceNodeName, sourcePortName, newState ) { 
- 
-   if ( arguments.length === 1 || arguments.length === 3 ) { 
-       if ( _.isUndefined( states[inportName] ) ) { 
-           return;
-       } 
-       return states[inportName];
-   }
+module.exports = function( data ) { 
 
-   if ( arguments.length === 2 || arguments.length === 4 ) { 
-       states[inportName] = { data: arguments[arguments.length-1],
-                              lm: Lm(),
-                              previousLms: [],
-                              port: inportName };
-       return this;
-   }
+       return { data: data, 
+                lm: Lm() };
 };
--- a/components/vni.js
+++ b/components/vni.js
@@ -12,13 +12,17 @@ vnis = [];
 // Returns the vni associated with the specified vnid.
 module.exports = function( vnid ) { 
  
-        if ( _.isUndefined( vnis[vnid] ) ) {
-           // Create Vni for real here
-        }
+    if ( _.isUndefined(vnid) ) { 
+        vnid='';
+    }
 
-        return {
-            inputState: inputStates,
-            outputState: outputStates.bind(this, 'output'),
-            errorState: outputStates.bind(this, 'error')
-        };
+    if ( _.isUndefined( vnis[vnid] ) ) {
+       // Create Vni for real here
+    }
+
+    return {
+        inputState: inputStates,
+        outputState: outputStates.bind(this, 'output'),
+        errorState: outputStates.bind(this, 'error')
+    };
 };
--- a/components/vni.js
+++ b/components/vni.js
@@ -2,11 +2,12 @@
 
 var _ = require('underscore');
 
-var outputStates = require('./outputStates');
 var inputStates = require('./inputStates');
+var outputStates = require('./outputStates');
 
 // Hash of vnis
 vnis = []; 
+
 
 // Returns the vni associated with the specified vnid.
 module.exports = function( vnid ) { 
@@ -17,7 +18,7 @@ module.exports = function( vnid ) {
 
         return {
             inputState: inputStates,
-            outputState: outputStates.bind( this, 'output' ),
-            errorState: outputStates.bind( this, 'error' ),
+            outputState: outputStates.bind(this, 'output'),
+            errorState: outputStates.bind(this, 'error')
         };
 };
--- a/components/vni.js
+++ b/components/vni.js
@@ -1,0 +1,23 @@
+// vni.js
+
+var _ = require('underscore');
+
+var outputStates = require('./outputStates');
+var inputStates = require('./inputStates');
+
+// Hash of vnis
+vnis = []; 
+
+// Returns the vni associated with the specified vnid.
+module.exports = function( vnid ) { 
+ 
+        if ( _.isUndefined( vnis[vnid] ) ) {
+           // Create Vni for real here
+        }
+
+        return {
+            inputState: inputStates,
+            outputState: outputStates.bind( this, 'output' ),
+            errorState: outputStates.bind( this, 'error' ),
+        };
+};
--- a/test/merge-patient-lab-iips-mocha.js
+++ b/test/merge-patient-lab-iips-mocha.js
@@ -5,151 +5,260 @@ var chaiAsPromised = require('chai-as-promised');
 chai.should();
 chai.use(chaiAsPromised);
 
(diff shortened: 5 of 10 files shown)


components/state.js
```diff
@@ -0,0 +1,26 @@
...
+
+var states = [];
+
+// Returns the vni associated with the specified vnid.
+module.exports = function( inportName, sourceNodeName, sourcePortName, newState ) { 
+ 
+   if ( arguments.length === 1 || arguments.length === 3 ) { 
+       if ( _.isUndefined( states[inportName] ) ) { 
+           return;
+       }
```
dbooth-boston: Are lines 13-15 needed, given that they return undefined anyway if states[inportName[ is undefined?

components/state.js
```diff
@@ -0,0 +1,26 @@
+// state.js
+
+var _ = require('underscore');
+
+var Lm = require('./Lm');
+
+var states = [];
```
dbooth-boston: It looks like you may be confusing State with InputStates.  AFAICT this module should only deal with a single State object, which has properties: data, lm, previousLms, oldLm, port.  See slide 9: http://dbooth.org/2013/pipeline/RDF-Pipeline-Implementation.pdf

components/state.js
```diff
@@ -0,0 +1,26 @@
...
+           return;
+       } 
+       return states[inportName];
+   }
+
+   if ( arguments.length === 2 || arguments.length === 4 ) { 
+       states[inportName] = { data: arguments[arguments.length-1],
+                              lm: Lm(),
+                              previousLms: [],
+                              port: inportName };
```
dbooth-boston: The port property should be the port object -- not its name.

components/state.js
dbooth-boston: Oh, this is a stub!  Sorry!

components/state.js
```diff
@@ -0,0 +1,26 @@
+// state.js
+
+var _ = require('underscore');
+
+var Lm = require('./Lm');
+
+var states = [];
+
+// Returns the vni associated with the specified vnid.
+module.exports = function( inportName, sourceNodeName, sourcePortName, newState ) {
```
dbooth-boston: I don't think the state module would have a function with that signature.  That looks more like a function for setting the inputStates object -- not a State object.  Is it written this way only for stub testing purposes?

test/merge-patient-lab-iips-mocha.js
```diff
@@ -5,151 +5,260 @@ var chaiAsPromised = require('chai-as-promised');
...
+
+                it("should have input state after input", function() {
+                    return Promise.resolve(componentFactory.getComponent )
+                        .then(commonTest.createComponent).then(function(component){
+
+                            // initialize state
+                            component.rpf.vni('').inputState('patient',
+                                                             { id: '001', 
+                                                               name: 'Alice', 
+                                                               dob: '1979-01-23' });
```
dbooth-boston: This last parameter should be a full State object (with data and lm properties)-- not just the data property value.

components/state.js
```diff
@@ -0,0 +1,26 @@
...
+       if ( _.isUndefined( states[inportName] ) ) { 
+           return;
+       } 
+       return states[inportName];
+   }
+
+   if ( arguments.length === 2 || arguments.length === 4 ) { 
+       states[inportName] = { data: arguments[arguments.length-1],
+                              lm: Lm(),
+                              previousLms: [],
```
dbooth-boston: I think previousLms should be undefined if it has not yet been set -- not an empty list, because an empty list would correspond to a node with no inputs (which is possible).

test/merge-patient-lab-iips-mocha.js
```diff
@@ -5,151 +5,260 @@ var chaiAsPromised = require('chai-as-promised');
...
+
+                            var state = component.rpf.vni('').inputState('patient');
+                            state.should.be.an('object');
+                            state.should.have.all.keys( 'data', 'lm', 'previousLms', 'port' );
```
dbooth-boston: An input connected to an IIP will not have a previousLms property.

test/merge-patient-lab-iips-mocha.js
```diff
@@ -5,151 +5,260 @@ var chaiAsPromised = require('chai-as-promised');
...
+                    });
+                });
+
+                it("should have input state data after input", function() {
+                    return Promise.resolve(componentFactory.getComponent )
+                        .then(commonTest.createComponent).then(function(component){
+
+                            // initialize state
+                            var patientState = { id: '001', name: 'Alice', dob: '1979-01-23' };
```
dbooth-boston: This is only the data part of a State object.  It needs to be a full State object, with data and lm properties.  The data property should hold the object on line196.

test/merge-patient-lab-iips-mocha.js
```diff
@@ -5,151 +5,260 @@ var chaiAsPromised = require('chai-as-promised');
...
+                it("should have a output state after input", function() {
+                    return Promise.resolve(componentFactory.getComponent )
+                        .then(commonTest.createComponent).then(function(component){
+
+                            // initialize state
+                            component.rpf.vni('').outputState( { id: '001', 
+                                                                 name: 'Alice', 
+                                                                 dob: '1979-01-23', 
+                                                                 glucose: '75',  
+                                                                 date: '2012-02-01'});
```
dbooth-boston: Again, this should be a full State object, with data and lm properties -- not just the data part.

test/merge-patient-lab-iips-mocha.js
```diff
@@ -5,151 +5,260 @@ var chaiAsPromised = require('chai-as-promised');
...
+
+                            var state = component.rpf.vni('').outputState();
+                            state.should.be.an('object');
+                            state.should.have.all.keys( 'data', 'lm', 'previousLms', 'port' );
```
dbooth-boston: No previousLms property when the input came from an IIP.

test/merge-patient-lab-iips-mocha.js
```diff
@@ -5,151 +5,260 @@ var chaiAsPromised = require('chai-as-promised');
...
+                });
+
+                it("should have an output state data after input", function() {
+                    return Promise.resolve(componentFactory.getComponent )
+                        .then(commonTest.createComponent).then(function(component){
+
+                            // initialize state
+                            var outputState = { id: '001', name: 'Alice', dob: '1979-01-23', 
+                                                glucose: '75',  date: '2012-02-01'};
+                            component.rpf.vni('').outputState( outputState );
```
dbooth-boston: Again, this needs to be a full State -- not just the data part.

test/merge-patient-lab-iips-mocha.js
```diff
@@ -5,151 +5,260 @@ var chaiAsPromised = require('chai-as-promised');
...
-                    labwork.send({id: '001',  glucose: '75',  date: '2012-02-01'});
-                    labwork.disconnect();
-                    component.inPorts.patient.detach(labwork);
-                    component.inPorts.patient.detach(patient);
+
+                }).then(commonStubs.promiseLater).then(function(component){
+
+                    return component.rpf.vni('').inputState('patient');
+
+                }).then(_.keys).then(_.sortBy).should.become(_.sortBy(['data','lm', 'previousLms', 'port']));
```
dbooth-boston: No previousLms when the input came from an IIP.

test/merge-patient-lab-iips-mocha.js
```diff
@@ -5,151 +5,260 @@ var chaiAsPromised = require('chai-as-promised');
...
                     return component;
-                }).then(promiseLater).then(function(component){
-                    return component.rpf.createVni('').getOutputState();
-                }).then(_.property('data')).should.become({id: '001',  name: 'Alice', dob: '1979-01-23', glucose: '75',  date: '2012-02-01'});
-            });
+
+                }).then(commonStubs.promiseLater).then(function(component){
+                    return component.rpf.vni('').inputState('patient');
+
+                }).then(_.property('data')).should.become({id: '001',  name: 'Alice', dob: '1979-01-23' });
```
dbooth-boston, line 205: Good

test/merge-patient-lab-iips-mocha.js
```diff
@@ -5,151 +5,260 @@ var chaiAsPromised = require('chai-as-promised');
...
+                    stubState( component );  // REMOVE WHEN STATES DONE
+                    commonTest.sendData( component, 'labwork',
+                                         {id: '001',  glucose: '75',  date: '2012-02-01'});
+                    return component;
+
+                }).then(commonStubs.promiseLater).then(function(component){
+
+                    return component.rpf.vni('').inputState('labwork');
+
+                }).then(_.keys).then(_.sortBy).should.become(_.sortBy(['data','lm', 'previousLms', 'port']));
```
dbooth-boston: No previousLms

test/merge-patient-lab-iips-mocha.js
```diff
@@ -5,151 +5,260 @@ var chaiAsPromised = require('chai-as-promised');
...
+
+                    stubState( component );  // REMOVE WHEN STATES DONE
+                    commonTest.sendData( component, 'labwork',
+                                         {id: '001',  glucose: '75',  date: '2012-02-01'});
+                    return component;
+
+                }).then(commonStubs.promiseLater).then(function(component){
+                    return component.rpf.vni('').inputState('labwork');
+
+                }).then(_.property('data')).should.become({id: '001',  glucose: '75',  date: '2012-02-01'});
```
dbooth-boston, line 236: good

test/merge-patient-lab-iips-mocha.js
```diff
@@ -5,151 +5,260 @@ var chaiAsPromised = require('chai-as-promised');
...
+        });
+
+        it.skip( "should have patient and labwork output state after input ports processing", function(done) {
+            return Promise.resolve(componentFactory.getComponent )
+                .then(commonTest.createComponent).then(function(component){
+
+                    sinon.stub(component.outPorts.output, 'send', function( data ) {
+                        data.should.exist;
+                        data.should.not.be.empty;
+                        data.should.have.ownProperty('output');
```
dbooth-boston: This test looks outdated.  It  looks partially like it is trying to test the return value of an updater function and partially like it is trying to test the value of the resulting output State object.

glenna-m: Note that it has the .skip extension.  That means that we are not executing it right now.  I carried it over so as not to lose it because having a test for these things is important and we don't have anything else that covers it. It's too hard to stub it so I have skip on it until the real branch.

components/state.js
glenna-m: This is not really part of this sprint or writing a behavioural unit test which is what I am trying to get and I don't have access to it with this level of the code. I will stub the port as null for now to make it clear it is not set.  I have to ask though - why do you want the port object itself on the inport state?

components/state.js
glenna-m: Broke out to an inportState and outportState to improve the signature.  Is that closer to what you want?

components/state.js
glenna-m: Removed it.

test/merge-patient-lab-iips-mocha.js
glenna-m: I think I need to understand why you want that passed in like that.   I have the fields that you document in the slides on the inputStates.    Why the extra layer?

test/merge-patient-lab-iips-mocha.js
glenna-m: The lm is being set inside the outputState function.    Not getting your issue here.  The output state structure that I am generating is: 
{ data: IIP data
 lm: lm }
and it is stored in an associative array indexed by output port as per spec.   Is there something more?

test/merge-patient-lab-iips-mocha.js
glenna-m: Same as above - the state generated has the lm and data.  We need to talk about this.

components/state.js
dbooth-boston: I'll have to see it.  Since this is a stub, I may not understand your intent.  If this were the real module for managing state objects, then I think it should only deal with state objects themselves -- not the inputStates object.  Parameters like sourceNodeName and sourcePortName only apply to the inputStates object.  I think the inputStates object should either be managed in the VNI module or in its own module, since a state object does not need to know about the inputStates object.

components/state.js
dbooth-boston: The port object is the port that owns that state.  Maybe it isn't needed.  I was thinking that it would allow the Framework to easily determine, for example, where messages should be sent if the state changes.  But since we do not have a known need for it yet, let's leave it out for now.  And maybe we'll find that we never need it.

test/merge-patient-lab-iips-mocha.js
dbooth-boston: The outputState function should be a getter/setter for an _entire_ state object -- not just the data property of that object.  It should not automatically set the lm property.  Generation of a new LM should be handled at a higher level in the Framework, because the lm should not always be changed just because the data property is changed.  Although it would be rare, a node should be able to update its state.data without updating its state.lm .

components/inputStates.js
```diff
@@ -0,0 +1,24 @@
+// inportStates.js
+
+var _ = require('underscore');
+var Lm = require('./Lm');
+
+var inportStates = [];
+
+// Returns the vni associated with the specified vnid.
```
dbooth-boston, line 8: Old comment?

components/inputStates.js
```diff
@@ -0,0 +1,24 @@
...
+module.exports = function( inportName, sourceNodeName, sourcePortName, newState ) { 
+ 
+   if ( arguments.length === 1 || arguments.length === 3 ) { 
+       if ( _.isUndefined( inportStates[inportName] ) ) { 
+           return;
+       }
```
dbooth-boston: Unless I'm missing something, I think lines 12-14 can be removed, since undefined would be returned anyway by line 15 if inportStates[inportName] is undefined.

components/outputStates.js
```diff
@@ -0,0 +1,23 @@
...
+
+var outportStates = [];
+
+// Returns the vni associated with the specified vnid.
+module.exports = function( outportName, newState ) { 
+ 
+   if ( arguments.length === 1 ) { 
+       if ( _.isUndefined( outportStates[outportName] ) ) { 
+           return;
+       }
```
dbooth-boston: Can lines 12-14 be deleted, since line 15 will return undefined anyway if outportStates[outportName] is undefined?

components/outputStates.js
```diff
@@ -0,0 +1,23 @@
...
+ 
+   if ( arguments.length === 1 ) { 
+       if ( _.isUndefined( outportStates[outportName] ) ) { 
+           return;
+       } 
+       return outportStates[outportName];
+   }
+
+   if ( arguments.length === 2 ) { 
+       outportStates[outportName] = { lm: Lm(),
```
dbooth-boston: newState should already be the whole state object

components/state.js
```diff
@@ -0,0 +1,26 @@
+// state.js
+
+var _ = require('underscore');
+
+var Lm = require('./Lm');
+
+var states = [];
+
+// Returns the vni associated with the specified vnid.
```
dbooth-boston, line 5: Old comment?

components/inputStates.js
glenna-m: Actually, it does work now - it must have been something else back when I was originally testing it.   Will remove it.

components/outputStates.js
```diff
@@ -0,0 +1,23 @@
+// outportStates.js
```
dbooth-boston: I think it makes sense to have an inputStates.js module, to manage the inputStates object, but it seems like the functionality in this module will be largely covered by in the states.js module.  Also, the output and error objects will be used differently by the Framework, so I think it would be simpler to just have separate output and error state objects directly in the VNI object -- no need for an array to hold them.  For these reasons I suggest dropping this module, and have its functionality covered by state.js and vni.js .

components/vni.js
```diff
@@ -0,0 +1,23 @@
+// vni.js
+
+var _ = require('underscore');
+
+var outputStates = require('./outputStates');
+var inputStates = require('./inputStates');
+
+// Hash of vnis
+vnis = [];
```
dbooth-boston, line 9: The vnis array should normally be stored in nodeInstance.rpf.vnis , so I don't know what the intent is here.  Is this just temporary for stub purposes?

test/merge-patient-lab-iips-mocha.js
dbooth-boston: The reason is that at the underlying Framework level, it must be possible to get/set an entire state object -- including data and lm properties -- not just the data part of it.  That is because an updater may need to be able to update the data part _without_ updating the lm.  It might do this, for example, if it decided to change the internal representation of the data but did not want to cause an event to be triggered notifying its downstream neighbors.

test/merge-patient-lab-iips-mocha.js
dbooth-boston: Got it. thanks

dbooth-boston:
Overall, I really like the structuring that you have done, of putting things into modules.  This is looking really good.

test/merge-patient-lab-iips-mocha.js
```diff
@@ -5,151 +5,260 @@ var chaiAsPromised = require('chai-as-promised');
...
-                    return component.rpf.createVni('').getInputState('patient', '', '');
-                }).should.become(undefined);
+
+        describe('#vni', function() {
+
+            it("should return a valid vni object", function() {
+                return Promise.resolve(componentFactory.getComponent )
+                  .then(commonTest.createComponent).then(function(component){
+
+                    var vni = component.rpf.vni('');
```
catch-point: Do we have to require the caller to include an empty string when asking for the IIP vni?

dbooth-boston: It's okay with me to have the vnid default to ''.  But as a hash index it must be '' -- not undefined -- because hash indexes are always converted to strings.

test/merge-patient-lab-iips-mocha.js
```diff
@@ -5,151 +5,260 @@ var chaiAsPromised = require('chai-as-promised');
...
+                it("should have input state after input", function() {
+                    return Promise.resolve(componentFactory.getComponent )
+                        .then(commonTest.createComponent).then(function(component){
+
+                            // initialize state
+                            component.rpf.vni('').inputState( 'patient',
+                                                              { id: '001', 
+                                                                name: 'Alice',
+                                                                dob: '1979-01-23' } );
+
```
dbooth-boston, line 104: This parameter should be a full state object -- not just the data part:
{ data: ... lm: ... }

test/merge-patient-lab-iips-mocha.js
```diff
@@ -5,151 +5,260 @@ var chaiAsPromised = require('chai-as-promised');
...
+                            var currentState = component.rpf.vni('').inputState('patient');
+                            currentState.should.be.an('object');
+                            currentState.should.have.all.keys( 'data', 'lm', 'port' );
```
dbooth-boston: We can leave off the 'port' property for this sprint, since we are not using it yet.

test/merge-patient-lab-iips-mocha.js
```diff
@@ -5,151 +5,273 @@ var chaiAsPromised = require('chai-as-promised');
...
 chai.use(chaiAsPromised);
 
+var sinon = require('sinon');
+
 var _ = require('underscore');
 var noflo = require('noflo');
+
+var componentFactory = require('../components/merge-patient-lab-iips');
 var getComponent = require('../components/merge-patient-lab-iips').getComponent;
+var State = require('../components/state');
```
catch-point, line 15: I don't think this test should call State() directly. Just in case there is a bug in there, I think the test should be explicit about the JSON object expected in the state and not use an external constructor (even if it is just a stub). We have to ensure the system creates the correct expected state explicitly and the only way to do that is to validate it against a different source. Same goes for Lm().

glenna-m, line 15: We can't have it both ways.  David is telling me that the state is the input to the inputState API.  You're telling me to just pas the values in there (as I originally did).  This may need a discussion.

components/outputStates.js
```diff
@@ -0,0 +1,20 @@
+// outputStates.js
```
dbooth-boston, line 1: Drop this?  Let's discuss

PR #22, opened 2016-02-05
Add test/merge-patient-lab-iips-mocha.js after discussion with the team

catch-point:


test/merge-patient-lab-iips-mocha.js
```diff
@@ -0,0 +1,283 @@
...
+chai.use(chaiAsPromised);
+
+var sinon = require('sinon');
+
+var _ = require('underscore');
+var noflo = require('noflo');
+
+var componentFactory = require('../components/merge-patient-lab-iips');
+var getComponent = require('../components/merge-patient-lab-iips').getComponent;
+var State = require('../components/state');
```
catch-point: I don't think this test should call State() directly. Just in case there is a bug in there, I think the test should be explicit about the JSON object expected in the state and not use an external constructor (even if it is just a stub). We have to ensure the system creates the correct expected state explicitly and the only way to do that is to validate it against a different source. Same goes for Lm().

PR #24, opened 2016-02-05
Event component factory

catch-point:
This gives the outPorts the same events as inPorts, which will allow both ports to be used the same way, regardless of the direction of the underlying socket. It will also give us a common way to register for event handlers, where they are noflo events or custom events, without having to pass around callback functions.


dbooth-boston:
LGTM

PR #26, opened 2016-02-05
Very simple create-state module and its unit tests for review

glenna-m:


components/State.js
```diff
@@ -0,0 +1,22 @@
+/**
+ *  file: State.js
```
dbooth-boston: good

test/State-mocha.js
```diff
@@ -0,0 +1,72 @@
...
+
+       // Set up a test state
+       var testState = State();
+
+        // Verify we got an object with the right keys
+       testState.should.be.an('object');
+       testState.should.have.all.keys('lm','data');
+
+       // Check the data content
+       expect( testState.data ).to.be.undefined;
```
dbooth-boston: perfect

test/State-mocha.js
```diff
@@ -0,0 +1,72 @@
...
+
+       // Check the LM
+       var components = testState.lm.match(/^LM(\d+)\.(\d+)$/);
+       components.should.have.length(3);
+    });
+
+    it('should use the lm parameter if it is defined', function() {
+
+       // Set up a test state
+       var testLm = Lm();
```
dbooth-boston: Slightly better to use a literal LM 'LM1328113669.00000000000000001' , as James pointed out earlier today

test/State-mocha.js
```diff
@@ -0,0 +1,72 @@
...
+
+        // Verify we got an object with the right keys
+       testState.should.be.an('object');
+       testState.should.have.all.keys('lm','data');
+
+       // Check the data content
+       testState.data.should.equal( testString );
+
+       // Check the LM
+       testState.lm.should.equal( testLm );
```
dbooth-boston: literal: 'LM1328113669.00000000000000001'

test/State-mocha.js
```diff
@@ -0,0 +1,72 @@
...
+       var testLm = Lm();
+       var testString = 'Some test data';
+       var testState = State( testString, testLm );
+
+        // Verify we got an object with the right keys
+       testState.should.be.an('object');
+       testState.should.have.all.keys('lm','data');
+
+       // Check the data content
+       testState.data.should.equal( testString );
```
dbooth-boston: literal would be slightly safer test: 'Some test data'

dbooth-boston:
Subject to minor notes, LGTM!

test/State-mocha.js
glenna-m: Will do

test/State-mocha.js
glenna-m: It is using a value.  Line above reads: 
    var testString='Some test data' 
That's what we are checking.    IMHO, hard coding it in the equals check does not make it safer in this case.  It just ensures we are duplicating the same literal two places.

This is not the same thing as the LM case James raised earlier.

dbooth-boston: Good point.  My mistake.

test/State-mocha.js
catch-point: A javascript string is immutable and cannot change it's state, whereas a javascript object is mutable and could be modified.

glenna-m: Were you suggesting a change here, James, or simply clarifying?

components/create-state.js
```diff
@@ -0,0 +1,28 @@
+/**
+ *  file: create-state.js
+ */
+
+var Lm = require('./Lm');
```
dbooth-boston, line 7: If we're switching to a lower case convention (as discussed with James today), maybe this should now be createLm.

components/create-state.js
```diff
@@ -0,0 +1,28 @@
...
+ * @param lm (optional) the last modified timestamp (Lm) for the object. 
+ *           See https://github.com/rdf-pipeline/framework/wiki/LM for more details.
+ *           If the lm is undefined, a new lm will be generated.
+ *
+ * @return the newly created state object.
+ */ 
+module.exports = function( vnid, data, lm ) { 
+
+       if ( arguments.length < 1 ) {
+           throw new Error("Unable to create state because no vnid was provided");
```
dbooth-boston, line 23: Good addition.  Sugggest slightly broader check instead: _.isUndefined(vnid)

test/create-state-mocha.js
```diff
@@ -0,0 +1,100 @@
...
+       state.should.have.all.keys('vnid', 'lm','data');
+
+       // Check the content
+       state.vnid.should.equal( testVnid );
+       expect( state.data ).to.be.undefined;
+
+       // Check the LM
+       var components = state.lm.match(/^LM(\d+)\.(\d+)$/);
+       components.should.have.length(3);
+    });
```
dbooth-boston, line 45: good

test/create-state-mocha.js
```diff
@@ -0,0 +1,100 @@
...
+var createState = require('../components/create-state');
+
+describe("State", function() {
+
+    it("should exist as a function", function() {
+        createState.should.exist;
+        createState.should.be.a('function');
+    });
+
+    it("should throw an error if no vnid specified", function() {
```
dbooth-boston, line 19: good

test/create-state-mocha.js
```diff
@@ -0,0 +1,100 @@
...
+       var testVnid = '123'; 
+       var testString = "Some test data";
+       var state = createState( testVnid, testString );
+
+        // Verify we got an object with the right keys
+       state.should.be.an('object');
+       state.should.have.all.keys('vnid', 'lm','data');
+
+       // Check the content
+       state.vnid.should.equal( testVnid );
```
dbooth-boston, line 79: good

dbooth-boston:
LGTM subject to minor comments noted

components/create-state.js
glenna-m, line 7: Yes, I considered that, but it will be for a separate check-in.  This is already the 4th review on what is essentially 7 lines of code.

dbooth-boston, line 7: ok

components/create-state.js
glenna-m, line 23: If vnid is be undefined, there won't be any arguments - the error will be thrown.  That won't make it safer logically.  I have  test that demonstrates this actually.

However, if you feel it is more clear, I can switch to that check.

dbooth-boston, line 23: I was thinking that checking _.isUndefined would also catch cases where a vnid parameter is passed, but for some reason it is undefined, due to some other mistake (like it wasn't set).

I could definitely imagine that happening.

catch-point:
LGTM

dbooth-boston:
LGTM

PR #27, opened 2016-02-06
adding InputStates module

glenna-m:


components/InputStates.js
```diff
@@ -0,0 +1,64 @@
...
+            return inputStates[inportName];
+        }
+
+        case requests.SET_IIP_STATE: {
+            inputStates[inportName] = arguments[arguments.length-1];
+            return this;
+        }
+
+        case requests.GET_EDGE_STATE: {
+            return inputStates[inportName][sourceNodeName][sourcePortName];
```
dbooth-boston: We previously planned to make inputStates be a hash of hash of hash, but since we changed previousLms to be an array of objects, I think we should do the same for inputStates, so that they have the same structure, as shown on slides 22-23: http://dbooth.org/2013/pipeline/RDF-Pipeline-Implementation.pdf
The reason is that previousLms will always be set from inputStates.  previousLms can be created for the first time by:

<pre>
if (!previousLms) {
  previousLms = _.map(inputStates, function(obj, i) { 
    return { inportName: obj.inportName, 
             sourceNodeName: obj.sourceNodeName, 
             sourcePortname: obj.sourcePortName } });
}
</pre>


And thereafter, whenever the output state changes, previousLms can be easily updated by:

<pre>
_.each(inputStates, function(obj, i) { previousLms[i].lm = obj.state.lm; } );
</pre>

components/InputStates.js
```diff
@@ -0,0 +1,64 @@
+/**
+ * InputStates.js
+ * 
+ * Manages the input states for a component's ports
+ */
+
+var _ = require('underscore');
+
+var inputStates = [];
```
dbooth-boston: JS noobie question here.  Does this create a singleton instance -- one per InputStates.js module?  It looks like it does.  But if so, then I don't understand how we will use this module to create one instance per VNI instance.  I've been searching around and have not been able to find an answer.  Maybe you can give me a quick explanation when you can.

components/InputStates.js
```diff
@@ -0,0 +1,64 @@
...
+ *
+ * @this noflo component context
+ * @return getter path returns the current input port state state.
+ *         setter path returns the current context 
+ */ 
+module.exports = function( inportName, sourceNodeName, sourcePortName, newState ) { 
+ 
+    switch (arguments.length) { 
+
+        case requests.GET_IIP_STATE: {
```
dbooth-boston, line 36: I think these would actually be slightly clearer if they were literal constants (1, 2, 3, etc.) with a comment saying "GET IIP state" etc., instead of symbolic constants, because in reading the code I had to look up the symbol definition to figure out the value.  However, if you prefer to do it this way it's no big deal.

test/InputStates-mocha.js
```diff
@@ -0,0 +1,261 @@
...
+        var testLm = "LM1328113669.00000000000000001";
+        var initialTestState = { data: testData, 
+                                 lm: testLm };
+        var inportName = "input";
+
+        // Set the input state for an "input" port
+        var context = InputStates( inportName, 
+                                   initialTestState );
+
+        // Context here will be the node.js context since mocha is running on that by default
```
dbooth-boston, line 56: Wouldn't the context normally be a VNI object, since that would be most convenient for chaining calls?   I would think it should be.  In which case I think it would be better if these tests set it up that way also.

It would be nice if this module did not have to deal with NoFlo components at all, since it only deals with RDF Pipeline objects.

glenna-m: There is no VNI object here because I deliberately broke this check in out from the vnis in hopes of actually getting something checked in and merged.   This check in is InputState only.  No VNI exists here. 

If we keep trying to get agreement on every part of the framework at every check in, I am concerned we will fail to make any forward progress.  When I check in the VNIs, by all means, you can request that the InputState should use VNI as the context.   If you believe VNIs should be checked in without InputState first, then let me know and I'll close this PR and do VNIs separately.

components/InputStates.js
glenna-m: There is no PreviousLMs on IIPs.  This check in is to support and complete our last sprint - which we still have not accomplished because we kept trying to bring in more than was required for it.

glenna-m: I am not creating PreviousLMs because they are not required for the IIP sprint.  If this is a general concern about the structure, then yes, I can make it an array instead of hash of hash of hash.  I did not realize you had changed this.

components/InputStates.js
glenna-m: Yes, this is singleton for this module instance.  If you want one per VNI, then you'll need to create one object per VNI.   Will see about that in the next pass of this.  But there is no VNI in this check in, you will note.

components/InputStates.js
glenna-m, line 36: Stylewise, I like the code to be self-documenting as much as possible; comments often get out of date, but the code does so less often.   With these names, it's pretty clear what those numbers represent.  Let's discuss this one with @jamesrdf  and make a decision between the 3 of us as to how to handle things like this so we are all on the same page.  Whatever is decided, I'll follow it.

test/InputStates-mocha.js
glenna-m, line 56: This code is not dealing with NoFlo, really.  It will return whatever context it was called with.  If you call it from a VNI and pass the VNI context, it will return the VNI context.  If you call it with NodeJS context, (and I have tests that do), it will return that context.  If you call it with a component, (and my tests do that tool), it will return the component context.   

There are no VNIs to test with for this check in.  That will be the next check in after we get this one approved and make progress.  I can certainly add a test showing you can get the VNI context when I do that.

components/InputStates.js
dbooth-boston: ok, thanks for educating me.

components/InputStates.js
dbooth-boston, line 36: sure, let's discuss it and choose.

components/InputStates.js
glenna-m: No worries.  And I have addressed the issue of easily integrating it into the vni in a check in I will make shortly.

components/InputStates.js
glenna-m, line 36: I did make some changes with comments that I think may help address your concern here too, since I think you did have a good point when I looked at it.  Will be in the next check in.

components/InputStates.js
dbooth-boston: Yes, i think should be an array, as shown in the slides, so that it will not have to be changed when previousLms is later implemented, and it will closely correspond.  thanks

test/InputStates-mocha.js
dbooth-boston, line 56: ok, thanks for clarifying.

test/InputStates-mocha.js
```diff
@@ -0,0 +1,261 @@
...
+        context.process.should.exist;
+        context.process.argv.should.exist;
+        context.process.env.should.exist;
+
+        // Verify the correct input state is there
+        var myInputState = InputStates( inportName );
+        myInputState.should.be.an('object');
+        myInputState.should.have.all.keys('data', 'lm');
+        myInputState.data.should.equal(testData);
+        myInputState.lm.should.equal(testLm);
```
dbooth-boston, line 67: good

components/InputStates.js
glenna-m: Will be in the next check in.

test/InputStates-mocha.js
```diff
@@ -0,0 +1,261 @@
...
+        context.process.env.should.exist;
+
+        // Verify the correct input state is there
+        var myInputState = InputStates( inportName,
+                                        sourceNodeName,
+                                        sourceNodePort );
+        myInputState.should.be.an('object');
+        myInputState.should.have.all.keys('data', 'lm');
+        myInputState.data.should.equal(testData);
+        myInputState.lm.should.equal(testLm);
```
dbooth-boston, line 100: good

test/InputStates-mocha.js
```diff
@@ -0,0 +1,261 @@
...
+                                                     inportName, 
+                                                     initialTestState );
+
+                // Verify context returned is the component context 
+                verifyContext( inputContext );
+
+                // Verify the correct input state for input port 
+                var myInputState = InputStates.call( component, 
+                                                     inportName );
+                verifyInputState( myInputState, testInportData, testInportLm );
```
dbooth-boston: good

test/InputStates-mocha.js
```diff
@@ -0,0 +1,261 @@
...
+
+                inputContext = InputStates.call( component, 
+                                                 inportName, 
+                                                 newTestState );
+
+                // Verify the input state for input port is now the new state 
+                var newInputState = InputStates.call( component, 
+                                                      inportName );
+                verifyInputState( newInputState, testInportData, testInportLm );
+                verifyInputState( newInputState, testInportData, testInportLm );
```
dbooth-boston: This line should be checking newTestState, shouldn't it?  Like this:

<pre>
verifyInputState( newTesttState, testInportData, testInportLm );
</pre>

test/InputStates-mocha.js
```diff
@@ -0,0 +1,261 @@
...
+                // Verify the correct input state for input port 1 
+                var myInputState = InputStates.call( component, 
+                                                     inportName,
+                                                     sourceNodeName,
+                                                     sourceNodePort );
+                verifyInputState( myInputState, testInportData, testInportLm );
+
+                // Verify the correct input state for input port 2  
+                // it should be undefined since we never set it
+                expect( InputStates('input2') ).to.be.undefined;
```
dbooth-boston, line 205: good

test/InputStates-mocha.js
```diff
@@ -0,0 +1,261 @@
...
+                // Verify context returned is the component context 
+                verifyContext( inputContext2 );
+
+                // Verify the correct input statefor input port 1 
+                var myInputState1 = InputStates( inportName1 );
+                verifyInputState( myInputState1, testInportData1, testInportLm1 );
+
+                // Verify the correct input statefor input port 2 
+                var myInputState2 = InputStates( inportName2 );
+                verifyInputState( myInputState2, testInportData2, testInportLm2 );
```
dbooth-boston, line 260: good

test/InputStates-mocha.js
glenna-m: No.  newTestState is the value we fed into the setter.  I want to test the state we got back from the getter after calling the setter - that is going to be the newInputState. 

Logically what I wrote is correct, but clearly it is confusing.  I'll adjust this to have less confusing variable names.

dbooth-boston: I wasn't confused by the variable names.  The two lines above look 
identical.  They seem to be testing the same thing twice.

components/InputStates.js
```diff
@@ -0,0 +1,85 @@
...
+                if ( inputState ) {
+                    return inputState.state;
+                }
+            } 
+            return;
+        }
+
+        // two arguments - inportName and newState.  Used for a constant input source.
+        case requests.SET_IIP_STATE: {
+            this.inputStates = this.inputStates || [];
```
dbooth-boston, line 49: move up

components/InputStates.js
```diff
@@ -0,0 +1,85 @@
...
+            this.inputStates.push({ inportName: inportName,
```
dbooth-boston, line 50: find also.  and insert in sorted order

PR #30, opened 2016-02-08
Restructure to new src & file naming conventions.

glenna-m:
Restructured code to move RDF framework modules to the src directory & updated all references.
Renamed Lm.js & its unit test to create-lm.js and create-lm-mocha.js as per new naming convention.


dbooth-boston:
LGTM

catch-point:
LGTM

PR #31, opened 2016-02-08
Add store-payload function

catch-point:
This function store the payload and returns all matching payloads (by vnid) from all the ports. Returns hash of port names to matching payload or array of payloads, if port is addressable, otherwise undefined, if no payload matches.

### DIFF
--- a/src/store-payload.js
+++ b/src/store-payload.js
@@ -1,0 +1,72 @@
+// store-payload.js
+
+var _ = require('underscore');
+var noflo = require('noflo');
+
+/**
+ * Stores the given payload for later use and returns a hash of the payloads by
+ * port names. If a vnid property is included in the payload, each payload with
+ * a unique vnid will be stored and only payloads with the same vnid will be
+ * returned. If a port is addressable then the payload from each socket is
+ * stored separately and an Array of payloads is returned for those ports.
+ *
+ * Usage:
+ *  ondata({vnid:'001',data:{},lm:'LM'}, 0) : {
+ *      input: {vnid:'001',data:{},lm:'LM'}
+ *  }
+ */
+module.exports = function(payload, socketIndex){
+    var vnid = _.has(payload, 'vnid') ? ('' + payload.vnid) : '';
+    storePayload(this, socketIndex, vnid, payload);
+    var inPorts = _.pick(this.nodeInstance.inPorts.ports, isPort);
+    var outPorts = _.pick(this.nodeInstance.outPorts.ports, isPort);
+    return _.defaults(
+        _.mapObject(inPorts, retrievePayload.bind(this, vnid)),
+        _.mapObject(outPorts, retrievePayload.bind(this, vnid))
+    );
+}
+
+/**
+ * Stores a payload on the given port by socketIndex and vnid.
+ * @param port the port to store the payload on
+ * @param socketIndex the index number of the socket for this payload or undefined
+ * @param vnid an identifier for the payload
+ */
+function storePayload(port, socketIndex, vnid, payload) {
+    var addressable = port.isAddressable();
+    var rpf = port.rpf = port.rpf || {};
+    if (addressable) {
+        rpf.payloads = rpf.payloads || [];
+        rpf.payloads[socketIndex] = rpf.payloads[socketIndex] || {};
+        rpf.payloads[socketIndex][vnid] = payload;
+    } else {
+        rpf.payloads = rpf.payloads || {};
+        rpf.payloads[vnid] = payload;
+    }
+}
+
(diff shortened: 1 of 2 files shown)


src/store-payload.js
```diff
@@ -0,0 +1,96 @@
...
+ */
+module.exports = function(port, vnid, payload, socketIndex){
+    if (!isPort(port)) throw Error("First argument must be a port");
+    if (!_.isString(vnid)) throw Error("Second argument must be a string");
+    if (arguments.length == 2) {
+        var inPorts = _.pick(port.nodeInstance.inPorts.ports, isPort);
+        var outPorts = _.pick(port.nodeInstance.outPorts.ports, isPort);
+        return _.defaults(
+            _.mapObject(inPorts, retrievePayload.bind(this, vnid)),
+            _.mapObject(outPorts, retrievePayload.bind(this, vnid))
```
dbooth-boston, line 33: Two other concerns: 1. I would rather not be creating more objects/arrays dynamically than are needed  If we can modify objects that are created once, that would reduce memory churn.  2. The previousLms are almost but not quite the same structure as inputStates, because they do not contain the states.

src/store-payload.js
```diff
@@ -0,0 +1,96 @@
...
+ *  ondata(new noflo.Port(), '001', {vnid:'001',data:{},lm:'LM'}, 0) : {
+ *      input: {vnid:'001',data:{},lm:'LM'}
+ *  }
+ */
+module.exports = function(port, vnid, payload, socketIndex){
+    if (!isPort(port)) throw Error("First argument must be a port");
+    if (!_.isString(vnid)) throw Error("Second argument must be a string");
+    if (arguments.length == 2) {
+        var inPorts = _.pick(port.nodeInstance.inPorts.ports, isPort);
+        var outPorts = _.pick(port.nodeInstance.outPorts.ports, isPort);
```
dbooth-boston: This creates arrays every time the function is called

src/store-payload.js
```diff
@@ -0,0 +1,96 @@
...
+        return this;
+    }
+};
+
+/**
+ * Stores a payload on the given port by socketIndex and vnid.
+ * @param port the port to store the payload on
+ * @param vnid a String to index the payload by
+ * @param payload the object to store or if undefined delete
+ * @param socketIndex the index number of the socket for this payload or undefined
```
dbooth-boston, line 46: Using socketIndex instead of <sourceNodeName, sourcePortName> seems like a good efficiency improvement.  Let's do that.

src/store-payload.js
```diff
@@ -0,0 +1,96 @@
...
+var _ = require('underscore');
+var noflo = require('noflo');
+
+/**
+ * Stores the given payload for later use or returns a hash of the payloads by
+ * port names. Each payload will have a unique index by vnid. If a port is
+ * addressable then the payload from each socket is stored separately and an
+ * Array of payloads is returned for those ports. Omitting the payload will
+ * not change anything, but return a hash of payloads by port name. Passing an
+ * undefined payload will remove any existing the indexed payload entry.
```
dbooth-boston, line 12: Overall, I like the idea of using socketIndex instead of sourceNodeName and sourcePortName to index the inputStates.  That seems easier and more efficient, and we can address the issue of making them predictable when the NoFlo graph is reloaded or reconfigured.  

However: 1. I would rather have the data stored on a per-VNI basis, instead of per-port.  I'm not seeing a real benefit to storing the data on the port, and it fragments the VNI object.  2. I would rather have the entire inputStates object be created at once (by createInputStates) when the VNI is created, rather than being created incrementally as events come in on different ports.  

Can you combine some of the ideas in this to the input-states.js module that Glenna drafted, to make a new input-states.js module?

src/store-payload.js
```diff
@@ -0,0 +1,96 @@
...
+ * Retrieves the payload (or Array of payloads if addressable) for the given
+ * vnid on the given port.
+ * @param vnid the vnid that the payload should have, if not the empty string
+ * @param port a port with the stored payloads
+ */
+function retrievePayload(vnid, port) {
+    if (!port.rpf) return;
+    var rpf = port.rpf;
+    if (port.isAddressable()) {
+        return _.map(rpf.payloads, function(payloads){
```
dbooth-boston, line 83: I would rather not be creating a new array every time the payload needs to be accessed

src/store-payload.js
catch-point, line 33: wrt 1. I think you may be under estimating the effectiveness of GC. GC is very efficient at removing recently created memory, besides this is a hash of port names, which will be very small. I don't understand your second concern.

PR #32, opened 2016-02-09
promise-output

catch-point:
Wraps a function and outputs the result to an output port and errors to an error port. This module also includes the default output/error port definitions.


dbooth-boston:
LGTM for now.  Issue 22 will be addressed in a later PR: https://github.com/rdf-pipeline/noflo-rdf-pipeline/issues/22

PR #33, opened 2016-02-16
Noflo component factory

catch-point:
Rename event-component-factory to noflo-component-factory


dbooth-boston:
LGTM

glenna-m:
The rename and changes to the test look good.  One thing I'd like to see added to the test suite would be something that verifies our ondata socketIndex handling is good.  It is currently working (I am using it), but I don't think we have any tests the ensure it stays good.

PR #34, opened 2016-02-16
New html-2-objects-compare.js component

glenna-m:
Generally useful component from demo that generates an html file comparing
two JavaScript objects side-by-side and opens them in the default browser.


components/html-2-objects-compare.js
```diff
@@ -0,0 +1,175 @@
...
+ * port will be shown in the right side column of the HTML page.  The title will be taken from the IIP 
+ * constant configuration for each port in the form of {"title":"<column title>"}.
+ */
+ 
+var _ = require('underscore');
+var util = require('util');
+var fs = require('fs');
+var open = require('open');
+
+var jswrapper = require('../src/javascript-wrapper');
```
catch-point, line 17: There is no javascript-wrapper in this branch :(

glenna-m, line 17: Correct.  As with the merge-patient code in sprint 5, javascript-wrapper is part of a separate check in.  This is simply to add a clean, very simple standard component to our toolbox of components.   It should run on whatever javascript-wrapper we decide to go with.   There is one available on the demo branch that could be used temporarily.   

I will be checking in a version of javascript-wrapper, framework, vni, and Input state later today(hopefully) in a separate pull request.  I expect that is going to involve a whole lot more discussion and changes which is why I kept this separate.

components/html-2-objects-compare.js
```diff
@@ -0,0 +1,175 @@
...
+    updater: function( file, left, right ) {   
+
+        // Verify that we have all data required to generate the html page. 
+        // TODO: Revisit and remove this check after more of the pipeline architecture is completed.
+        if ( shouldFireUpdater( file, left, right) ) { 
+
+            // We have both an IIP and port data on these ports.  It can theoretically arrive in any order.
+            // sortPortData return an object with a title and a data field so we know what we are referencing
+            // henceforth.
+            var leftOutput = sortPortData( left, 'left' );
```
catch-point, line 68: Why is the title sent to the same port as the json object?

components/html-2-objects-compare.js
```diff
@@ -0,0 +1,175 @@
...
+                          "           <th bgcolor='#c8c8c8' width='50%'> " + rightOutput.title + " </th>" + 
+                          "         </tr>\n" +
+                          "         <tr>\n" + 
+                          "           <td bgcolor='#e8e8e8'> <pre> "+ util.inspect( leftOutput.data, {depth: null} ) + "</pre></td>" +
+                          "           <td> <pre> "+ util.inspect( rightOutput.data, {depth: null} ) + "</pre></td>";
+                          "         </tr>\n" +
+                          "      </table>\n" +
+                          "  </body>\n" +
+                          "</html>";
+            fs.writeFileSync(file, content);
```
catch-point, line 88: In the spirit of doing only one thing, what about just producing the content and letting a down stream component determine what to do with it?

components/html-2-objects-compare.js
glenna-m, line 68: Because the title is what will be displayed as the column title of that json object.  I thought it made sense to keep related things together.  I also chose to do it this way so we had a test component that had both IIP and packet input on the same port since it's a scenario we will need to be testing in the framework.

This component will be displaying one object on the left (with the IIP title specified for that port/object,and one object on the right (with it's own IIP title).

dbooth-boston:
LGTM

components/html-2-objects-compare.js
glenna-m, line 88: The question of what's the right component granularity in Noflo is always an interesting one with advantages and disadvantages.    In this case, the advantage to splitting it is flexibility - you could feed a stream or a file or something else.   The disadvantage is simplicity in a demo (more clutter in the graphs we show), ease of testing (it's easy to diff files) and ease of setting up a demo (more  components to find and configure instead of one).

I'm open to splitting it up in the future if we find a use case where that would be helpful to us.    For now, however, given it's not a lot of code and the idea of generating an html file page with a table is pretty simple, I'd rather merge as is so we have it available, and revisit improving it if/when we have a need to do it.   Sound reasonable?

components/html-2-objects-compare.js
catch-point, line 68: What about hard coding the titles to "Before"/"After" and renaming the ports? I think this would make the component more intuitive to use.

components/html-2-objects-compare.js
catch-point, line 88: okay

components/html-2-objects-compare.js
dbooth-boston, line 68: But it may be used for an A/B comparison instead of before/after.

dbooth-boston:
Since this is not one of our important deliverables I don't think we should spend more time on it.  Let's just get it merged and move on.  thanks

PR #35, opened 2016-02-16
Noflo access

catch-point:
This change noflo-component-factory to use facade objects as the context to event handlers


src/noflo-component-access.js
```diff
@@ -0,0 +1,138 @@
...
+function facadePort(port) {
+    return _.extend({
+        // port.nodeInstance might not be assigned yet
+        get nodeInstance() {
+            return port.nodeInstance && facadeComponent(port.nodeInstance)
+        },
+        get name() {
+            return port.name;
+        }
+    }, facadeFunctions(port,
```
glenna-m, line 38: Might be good to add a comment with a link to the noflo documentation that describes the attribute associated with these functions.

src/noflo-component-access.js
```diff
@@ -0,0 +1,138 @@
...
+}
+
+/**
+ * Picks all the given functions from object and returns the set bounded to the
+ * original object context. If any argument is passed to those functions that is
+ * also a function it is bound to the called context.
+ * @param object that contains functions
+ * @param keys explicit names of functions to include in result
+ */
+function facadeFunctions(object /* keys */) {
```
glenna-m, line 72: If I'm reading this right, the object is not really an object, is it?  It's more of a list of function names as keys, no?

src/noflo-component-access.js
```diff
@@ -0,0 +1,138 @@
...
+ */
+function facadeFunction(fn) {
+    if (!_.isFunction(fn))
+        return fn;
+    else if (_.isFunction(fn._bind_facade_noflo_access))
+        return fn._bind_facade_noflo_access;
+    else
+        return fn._bind_facade_noflo_access = fn.bind(this);
+}
+
```
glenna-m, line 107: I kind of like the elegance of your applyFacade and facadeFunction.  Very flexible.

src/noflo-component-access.js
catch-point, line 72: object is a noflo.Component, noflo.OutPort, or noflo.InPort instance in this case

catch-point:
setMaxListeners is included bc it is part of the EventEmitter[1] API. It is a common API and there is no reason not to include all ten functions.

[1] https://nodejs.org/dist/latest-v4.x/docs/api/events.html#events_emitter_setmaxlisteners_n

src/noflo-component-access.js
glenna-m, line 72: Could line 88 be updated to note that?

glenna-m:
Would still like to see a couple comments as a kindness to future readers of this code.   Once that's done, I am comfortable with the code and can give it an LGTM, pending @dbooth-boston  approval.

src/noflo-component-access.js
catch-point, line 38: Can you give an example?

src/noflo-component-access.js
```diff
@@ -0,0 +1,143 @@
...
+/**
+ * Create a noflo Port facade from a given noflo Port that includes only a limited
+ * subset of properties and functions.
+ * This facade mimic the main usage of http://noflojs.org/api/OutPort.html
+ */
+function facadePort(port) {
+    return _.extend({
+        // port.nodeInstance might not be assigned yet
+        get nodeInstance() {
+            return port.nodeInstance && facadeComponent(port.nodeInstance)
```
dbooth-boston: Won't this be creating a new facadeComponent every time port.nodeInstance is accessed at runtime?  I think the facade should only be created once for each component instance.

src/noflo-component-access.js
```diff
@@ -0,0 +1,143 @@
...
+        },
+        get outPorts() {
+            return _.mapObject(_.pick(component.outPorts, isOutPort), facadePort);
+        },
+        get rpf() {
+            return component.rpf = component.rpf || {};
+        }
+    }, facadeFunctions(component,
+        'getDescription',
+        'getIcon'
```
dbooth-boston: If the RDF Pipeline Framework does not actually use these functions (getDescription and getIcon) then I would rather not expose them in the facade.

src/noflo-component-access.js
```diff
@@ -0,0 +1,143 @@
...
+        'isBuffered',
+        'isRequired',
+        'isCaching',
+        'isConnected',
+        'connect',
+        'beginGroup',
+        'send',
+        'endGroup',
+        'disconnect',
+        'checkRequired'
```
dbooth-boston: Is this exposing all of these functions in the facade?  We should minimize the number of functions that we expose in the facade: only expose the ones we really need.

src/noflo-component-access.js
```diff
@@ -0,0 +1,143 @@
...
+        'addListener',
+        'emit',
+        'getMaxListeners',
+        'listenerCount',
+        'listeners',
+        'on',
+        'once',
+        'removeAllListeners',
+        'removeListener',
+        'setMaxListeners'
```
dbooth-boston: Again, we should minimize the number of NoFlo functions that the facade exposes.

src/noflo-component-access.js
```diff
@@ -0,0 +1,143 @@
...
+
+/**
+ * Picks all the given functions from object and returns the set bounded to the
+ * original object context. If any argument is passed to those functions that is
+ * also a function it is bound to the called context.
+ * @param object with functions, i.e. noflo.Component, noflo.InPort/OutPort
+ * @param keys explicit names of functions to include in result
+ */
+function facadeFunctions(object /* keys */) {
+    return _.mapObject(_.pick(object, arguments), function(fn) {
```
dbooth-boston: Should that be the following, to avoid trying to pick 'object' as a key?   Although maybe it's harmless to try to pick 'object' also.

<pre>
return _.mapObject(_.pick(object, _.toArray(arguments).slice(1)), function(fn) {
</pre>

dbooth-boston:
We need to discuss this.  If I've understood the code correctly, it looks like it defines general purpose functions that generate facade functions that will map to their corresponding NoFlo implementations.

src/noflo-component-access.js
catch-point: These functions are not part of noflo, but part of nodejs and fairly commonly used in nodejs dev.

src/noflo-component-access.js
catch-point: Good catch

src/noflo-component-access.js
catch-point: Unlike in Java, objects can change and the best way to ensure the caller doesn't modify the facade is to replace it often. However, it make it behave more like noflo I have now changed to to reuse facade and preserve any changes on it.

src/noflo-component-access.js
catch-point: Removed

src/noflo-component-access.js
catch-point: I removed getDescription, isBuffered and isCaching, anything else?

dbooth-boston:
LGTM, subject to changes discussed with James on Friday (mostly removing functions that are not needed by the framework).

src/noflo-component-access.js
```diff
@@ -0,0 +1,121 @@
...
+ * Create a noflo Port facade from a given noflo Port that includes only a limited
+ * subset of properties and functions.
+ * This facade mimic the main usage of http://noflojs.org/api/OutPort.html
+ */
+function facadePort(nodeInstance, port, name) {
+    return _.extend({
+        name: name,
+        nodeInstance: nodeInstance
+    }, facadeFunctions(port,
+        'isAddressable',
```
glenna-m, line 39: isMulti?

test/noflo-component-factory-mocha.js
```diff
@@ -32,11 +32,114 @@ describe('noflo-component-factory', function() {
...
+                    input2:{
+                        ondata: function(payload) {
+                            this.on('change', handler);
+                            this.emit('change', payload);
+                        }
+                    }
+                }
+            }));
+            var input1 = noflo.internalSocket.createSocket();
+            component.inPorts.input1.attach(input1);
```
glenna-m: Can you use the common-test.sendData function so we don't keep duplicating all this code?  It's less to write and read.

test/noflo-component-factory-mocha.js
```diff
@@ -32,11 +32,114 @@ describe('noflo-component-factory', function() {
...
+            var component = test.createComponent(componentFactory({
+                inPorts:{
+                    input:{
+                        ondata: function(payload) {
+                            done(this.name);
+                        }
+                    }
+                }
+            }));
+            var socket = noflo.internalSocket.createSocket();
```
glenna-m: Another place for common-test.sendData

test/noflo-component-factory-mocha.js
```diff
@@ -32,11 +32,114 @@ describe('noflo-component-factory', function() {
...
+                inPorts:{
+                    input:{
+                        description: "World",
+                        ondata: function(payload) {
+                            done(payload + ' ' + this.isAddressable());
+                        }
+                    }
+                }
+            }));
+            var socket = noflo.internalSocket.createSocket();
```
glenna-m: Another common-test.sendData

test/noflo-component-factory-mocha.js
```diff
@@ -32,11 +32,114 @@ describe('noflo-component-factory', function() {
...
+                    'input2':{
+                        ondata: function(payload) {
+                            if (this.nodeInstance.payload)
+                                done(this.nodeInstance.payload + ' ' + payload);
+                            else this.nodeInstance.payload = payload;
+                        }
+                    }
+                }
+            }));
+            var input1 = noflo.internalSocket.createSocket();
```
glenna-m: Another sendData place

glenna-m:
I think this code is ready to merge after modifying the tests to use the common-test.sendData API instead of duplicating everything.

PR #36, opened 2016-02-18
Adding RDF Pipeline input-states module

glenna-m:
Latest round of input-states module

### DIFF
--- a/src/input-states.js
+++ b/src/input-states.js
@@ -1,0 +1,220 @@
+/**
+ * input-states.js
+ * 
+ * Manages RDF pipeline input states for a VNI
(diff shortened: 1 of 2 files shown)


dbooth-boston:
Very clear and easy to read.  Thanks!

LGTM

catch-point:
This doesn't look like the API we discussed.

glenna-m:
As I worked with David's design, I found we needed a number of changes to make it work more cleanly.  I am going to check in the full set of javascript-wrapper, framework, input-states in a separate PR, since I do have an end to end flow now.  I expect we'll need a lot of discussion and probably some refinement on it.

PR #37, opened 2016-02-19
Add round-robin test and rename ports to input/output

catch-point:


test/round-robin-mocha.js
```diff
@@ -0,0 +1,89 @@
...
+ * Creates and starts a noflo.Network with a component for every component module
+ * given, however, no edges are present.
+ * Usage:
+ *  createNetwork({name:require('../components/rdf')}).then(function(network){
+ *      network.processes.name.component is the component instance
+ *      network.graph.addEdge('name', 'output', 'name', 'input') to add edge
+ *      network.graph.addInitial(data, 'name', 'input') to send data
+ *  });
+ */
+function createNetwork(componentModules) {
```
dbooth-boston, line 68: This looks like something we could use more broadly in our test tool kit.  Should this be included in common-test.js or common-stubs.js?

dbooth-boston:
LGTM, but note my comment above.

test/round-robin-mocha.js
catch-point, line 68: yes, I think it is easier to move to common-test once it is used in develop branch

PR #38, opened 2016-02-19
Use promise-output in request-template

catch-point:
Replace promise-component with promise-output in request template


test/request-template-mocha.js
```diff
@@ -0,0 +1,145 @@
...
+ * Usage:
+ *  createNetwork({name:require('../components/rdf')}).then(function(network){
+ *      network.processes.name.component is the component instance
+ *      network.graph.addEdge('name', 'output', 'name', 'input') to add edge
+ *      network.graph.addInitial(data, 'name', 'input') to send data
+ *  });
+ */
+function createNetwork(componentModules) {
+    var graph = new noflo.Graph();
+    _.each(componentModules, function(module, name) {
```
dbooth-boston, line 132: I'm fine with this being written this way, but I am curious to understand why it is.  Is this written using _.each (instead of 'for (name in componentModules)') because of JavaScript's lack of local variables?   Or because of the need to otherwise check for 'own' properties?  Or both?

dbooth-boston:
LGTM.  We should migrate to higher level functions for creating our test networks as soon as we can, but this looks good as a current step along the way.

glenna-m:
James will need to weigh in, but the reason that I often use forEach or _.each instead of a for loop is to minimize the use of mutable variables.  One of the best practices that has come into popular acceptance  over the last few years is the idea that mutable variables are more often a source of bugs than non-mutable variables since programmers usually get it right in the beginning, but screw up later with unforeseen side effects.   

The other reason for using some of the underscore modules would be to benefit from any functional programming in the implementation which tries to avoid processing anything more than it must.  Javascript is not a functional programming language, of course, but underscore is one of the libraries that tries to move it that direction.

catch-point:
Glenna is correct, debugging code is hard enough without temporal states. I tried to avoid re-assignment of variables (and state mutation) as much as possible.

dbooth-boston:
Thanks!  That's helpful to know.

PR #39, opened 2016-02-19
Use promise-output in rdf components

catch-point:
Replace promise-component usage with promise-output in rdf components


dbooth-boston:
LGTM.  Again, we should move to using the higher-level RDF Pipeline functionality ASAP (since this is still using noflo functionality directly if I've understood correctly), but this looks like a great step until then.

PR #40, opened 2016-02-19
Use promise-output in http-accept

catch-point:
Replace promise-component with promise-output in http-accept component


dbooth-boston:
LGTM

PR #41, opened 2016-02-19
Remove promise-component

catch-point:
Replace all usages of promise-component with promise-output


dbooth-boston:
LGTM

components/http-accept.js
```diff
@@ -19,35 +19,43 @@ exports.getComponent = promiseComponent({
...
         }
-    },
+    }, promiseOutput.outPorts),
     inPorts: {
         limit: {
             description: "Request Content-Length limit",
             datatype: 'int',
-            ondata: basenode.assign('limit')
+            ondata: function(limit) {
+                this.nodeInstance.limit = limit;
```
glenna-m, line 28: Until the discussion this morning, I would have been concerned this should be this.nodeInstance.rpf.  Will this will now be assigned to the noflo facade?

components/rdf-load.js
```diff
@@ -4,48 +4,58 @@ var _ = require('underscore');
...
-            ondisconnect: load
+            ondata: function(data) {
+                this.nodeInstance.data = data;
+            },
+            ondisconnect: promiseOutput(load)
         }
     }
 });
 
 function load() {
```
glenna-m, line 55: Some of this function should probably be in the new noflo access layer, yes?   For instance, graphs and options are pretty noflo-specific for a piplein component.   I realize that will be on one of the other PRs.  Just checking that you also believe it should move and we will be moving it.

components/rdf-query.js
```diff
@@ -5,50 +5,62 @@ var Promise = require('promise');
...
             datatype: 'object',
             required: true,
-            ondata: execute
+            ondata: promiseOutput(execute)
         }
     }
 });
 
 
 function execute(graph) {
```
glenna-m, line 64: Can we get standard and add comments on major functions like this?

package.json
```diff
@@ -22,9 +22,13 @@
       "url": "http://opensource.org/licenses/apache2.0.php"
     }
   ],
-  "dependencies": {
+  "devDependencies": {
```
glenna-m, line 25: I like breaking this out.  Nice.

glenna-m:
I know you deleted the *.json graphs because the new tests generate the graphs on the fly - very nice work.  But it is sometimes nice to have the graphs there so one can load them in noflo and see what these components are configured for and doing interactively.  Are we sure it's a good idea to delete them?  This is a topic for discussion @dbooth-boston @jamesrdf

components/http-accept.js
catch-point, line 28: Yes, with noflo-access branch merged in, this would be assigned to the component facade

test/rdf-components-mocha.js
```diff
@@ -0,0 +1,139 @@
+// rdf-components-mocha.js
+
```
glenna-m, line 2: Good tests for the cases that all succeed.  Would be good to add some tests where things go wrong e.g., query or update fails, so we can verify error handling.  This does not block merging this down - it's a suggestion for future improvement next time we are in here.

components/rdf-load.js
catch-point, line 55: I don't think the load function should be moved anywhere else. The term graph here refers to an RDF graph and the options object is for use with the rdfstore. Neither of which are about noflo.

test/request-template-mocha.js
```diff
@@ -0,0 +1,151 @@
...
+ * Creates and starts a noflo.Network with a component for every component module
+ * given, however, no edges are present.
+ * Usage:
+ *  createNetwork({name:require('../components/rdf')}).then(function(network){
+ *      network.processes.name.component is the component instance
+ *      network.graph.addEdge('name', 'output', 'name', 'input') to add edge
+ *      network.graph.addInitial(data, 'name', 'input') to send data
+ *  });
+ */
+function createNetwork(componentModules) {
```
glenna-m: How many createNetwork functions do we need?  Can you move this to a common-test function?

glenna-m:
None of my comments on this PR are show stoppers.  It would be nice to do a little clean up adding the comments and removing duplicate code, but if it will take too long, it's OK to merge this so we make some progress.

components/rdf-load.js
glenna-m, line 55: OK - can you call it rdfGraph & rdfOptions?  Since the terms are also used in noflo, it would be good to make it obvious what you are working with.

PR #43, opened 2016-02-20
End-to-end flow for Sprint 5

glenna-m:
This includes input-states, vnis, framework, and javascript-wrapper

A end-to-end Sprint 5 Implementation
This is my cut at an end-to-end Sprint 5 implementation, incorporating a number of the most recent design ideas our team has been discussing. I have a working demo to show on Monday of Sprint 5.

As James noted with the input-states, I have diverged somewhat from the original behavioural test suite that we wrote. I typically did so for one of two reasons:
1. I wanted to bring the design into alignment with some of David's recent design thinking and our team's discussions.
2. The original interface was a bit awkward and I felt it would be improved with some change. I spent a whole lot more time thinking about how this should all flow cleanly while implementing and testing this code than we did during the hour or two when we first wrote that behavioral test. I knowingly did take some liberties here, knowing that we'd have to discuss them before it would be approved. And if we feel they aren't the right call, I will adjust as needed based on team discussion.

I don't expect this will be approved quickly given our team's history, but, I do think this is the closest we've ever had to something approximating David's RDF Pipeline flow. I'm hoping we can use this as a basis for discussion on Monday.

### DIFF
--- a/src/framework.js
+++ b/src/framework.js
@@ -1,0 +1,194 @@
+/**
+ * File: framework.js 
+ */
+var _ = require('underscore');
+var util = require('util');
+
+var stateFactory = require('./create-state');
+var vnis = require('./vnis');
+
+/**
+ * This module implements the core of the RDF Pipeline Framework.  
+ */
+module.exports = {
+
+    componentFactory: function( nodeDef, wrapper ) { 
+        
+        var compFactory = require('./noflo-component-factory');
+        return compFactory(
+                _.defaults( {
+                     // Add an ondata function to be invoked every time we receive new data
+                     inPorts: _.mapObject(nodeDef.inPorts, function( port, portName ) {  
+                         return _.extend({
+                             ondata: function(payload, socketIndex) {
+                                   var that = this; 
+                                   return ondata.call( that, payload, portName, socketIndex, port );
+                             }
+                         }, port);
+                     })
+                 }, nodeDef ),
+                { rpf: {      // extend default noflo component node instance with framework rpf object 
+                     vnis: vnis,
+                     wrapper: wrapper
+                  }
+                } 
+        );
+    }
+
+}; // module.exports
+
+
+var ondata = function( payload, portName, socketIndex, port ) {
+
+     // socketIndex is only defined for addressable/multi ports so set default of 0 if not defined.  
+     // If there is only one input, we can just default to socket index 0, so if not defined, 
+     // do this for now.   We need to discuss whether addressable should just be required on our 
+     // pipeline ports.  IMHO, I think it probably should be to minimize potential user-error in this area
+     // and simplify the interface our users work with.
+     socketIndex = socketIndex || 0;
+
+     var vnid = payload.vnid || '';
+     var vni = vnis.get.call( this.nodeInstance, vnid );
+     vni.inputStates.set.call( vni.node, 
+                               portInfo.call( vni.node, portName, socketIndex ),
+                               stateFactory( vnid, payload ) );
+
+     // check if should run updater & run if necessary
+     if ( shouldRunUpdater( vni ) ) { 
+        var data = this.nodeInstance.rpf.wrapper.fRunUpdater( vni.inputStates );
+        if ( data ) {
+            var outputState = stateFactory( vnid, data );
+            vni.outputState = outputState;
+
+            // TODO: Discuss whether this should be in the access layer instead of here
+            this.nodeInstance.outPorts.output.send( outputState )
+            this.nodeInstance.outPorts.output.disconnect();
+
+            // Clean up input states
+            vni.inputStates.deleteAll();
+        }
+     }
+
+     return;
+};
+
+/** 
+ * Check the input to see if we have all data required to run the updater or not. 
+ * This is simply a default updater policy - we may have other policies in the future.
+ * 
+ * @param vni virtual node instance whose input states are to be checked.
+ *
+ * @return true if we have received data on all required input port edges
+ */
+function shouldRunUpdater( vni ) { 
+
+    var requiredInPorts = requiredInputPorts.call( vni.node );
+
+    if ( ! _.isEmpty( requiredInPorts ) ) { 
+
+        // Check if we have input on all the required input ports and their edges
+        for ( var i=0, max=requiredInPorts.length; i < max; i++ ) { 
+            var port = requiredInPorts[i];
+            if ( ! hasAllInputs( port, vni.inputStates.count( port.name ))) {
+                return false;
+            }
+        }
+    }
+ 
+    return true;
+}
+
+/**********************************************************************************/
+/* Move the following functions to the noflo access layer as soon as we have one! */
+/**********************************************************************************/
+
+/**
+ * Check to see if we have received input on all of a port's edges or not.
+ * TODO: Move this to the noflo access layer
+ * 
+ * @param port a noflo port object
+ * @param inputCount the number of inputs received for this port so far.
+ *
+ * @return true if the inputCount matches the number of sockets for the port.
+ *         There will be one socket for each connected input edge.  
+ */
+function hasAllInputs( port, inputCount ) {
+    return (port.sockets.length === inputCount); 
+}
+
+/** 
+ * Examines the port socket to determine if the current data input is for an IIP input.
+ *
+ * TODO: Move this to the noflo access layer
+ * TODO: Discuss how to handle this if we do not have an addressable port - I don't think it's possible 
+ *       to tell if it's an IIP input if there are multiple inputs and the port is not addressable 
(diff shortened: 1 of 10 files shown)


src/framework.js
```diff
@@ -0,0 +1,194 @@
...
+    componentFactory: function( nodeDef, wrapper ) { 
+        
+        var compFactory = require('./noflo-component-factory');
+        return compFactory(
+                _.defaults( {
+                     // Add an ondata function to be invoked every time we receive new data
+                     inPorts: _.mapObject(nodeDef.inPorts, function( port, portName ) {  
+                         return _.extend({
+                             ondata: function(payload, socketIndex) {
+                                   var that = this;
```
dbooth-boston, line 24: JS noobie question:  Is the line above actually needed in this case?  Since no function is being returned in the next line, could the next line have been written as follows instead?

<pre>
return ondata.call( this, payload, portName, socketIndex, port );
</pre>

src/framework.js
```diff
@@ -0,0 +1,194 @@
...
+                     wrapper: wrapper
+                  }
+                } 
+        );
+    }
+
+}; // module.exports
+
+
+var ondata = function( payload, portName, socketIndex, port ) {
```
dbooth-boston, line 41: Why is port passed as an explicit parameter?  I thought that 'this' would be the port when ondata is called.  Did that change?  

[Added later]
Actually, from reading line 21 above it looks like the 'port' parameter actually holds the port definition -- not the runtime port object (unless they're the same thing?).  If that's correct then it may be clearer to call it portDef.

src/framework.js
```diff
@@ -0,0 +1,194 @@
...
+
+     // socketIndex is only defined for addressable/multi ports so set default of 0 if not defined.  
+     // If there is only one input, we can just default to socket index 0, so if not defined, 
+     // do this for now.   We need to discuss whether addressable should just be required on our 
+     // pipeline ports.  IMHO, I think it probably should be to minimize potential user-error in this area
+     // and simplify the interface our users work with.
```
dbooth-boston, line 47: The policy that I planned was: 1. users must set the 'multi' option to true if they want to attach more than one thing to the port; and 2. an array of data objects (probably {sourceNodeName: ssn, sourcePortName: spn, data: data} objects) is always passed to each updater parameter whose port was declared 'multi'.  Some consequences of this policy: (a) it's an error if more than one thing is attached to a node that was not declared 'multi'; (b) if a port is 'required' and 'multi' then it is an error if it has zero things attached; (c) if a port is 'multi' and not 'required' and has zero things attached then its corresponding updater parameter receives an array with zero elements.

src/framework.js
```diff
@@ -0,0 +1,194 @@
...
+                               stateFactory( vnid, payload ) );
+
+     // check if should run updater & run if necessary
+     if ( shouldRunUpdater( vni ) ) { 
+        var data = this.nodeInstance.rpf.wrapper.fRunUpdater( vni.inputStates );
+        if ( data ) {
+            var outputState = stateFactory( vnid, data );
+            vni.outputState = outputState;
+
+            // TODO: Discuss whether this should be in the access layer instead of here
```
dbooth-boston, line 63: Yes, the access layer should provide a sendState function.  (James and I on Friday discussed calling it sendData, but, on thinking more about it, since it will be sending a state object it might as well be called sendState.)

src/framework.js
```diff
@@ -0,0 +1,194 @@
...
+
+var ondata = function( payload, portName, socketIndex, port ) {
+
+     // socketIndex is only defined for addressable/multi ports so set default of 0 if not defined.  
+     // If there is only one input, we can just default to socket index 0, so if not defined, 
+     // do this for now.   We need to discuss whether addressable should just be required on our 
+     // pipeline ports.  IMHO, I think it probably should be to minimize potential user-error in this area
+     // and simplify the interface our users work with.
+     socketIndex = socketIndex || 0;
+
```
dbooth-boston, line 49: Normalize payload first, i.e., if it is an IIP payload then wrap a new state around it:

<pre>
if (_.isUndefined(payload.vnid)) {
  // IIP payload
  payload = stateFactory('', payload);
}
</pre>

src/framework.js
```diff
@@ -0,0 +1,194 @@
...
+     var vnid = payload.vnid || '';
+     var vni = vnis.get.call( this.nodeInstance, vnid );
+     vni.inputStates.set.call( vni.node, 
+                               portInfo.call( vni.node, portName, socketIndex ),
+                               stateFactory( vnid, payload ) );
```
dbooth-boston, line 54: This line should just pass the payload (which is now always a state already) instead of creating a new state.

src/framework.js
```diff
@@ -0,0 +1,194 @@
...
+
+     // check if should run updater & run if necessary
+     if ( shouldRunUpdater( vni ) ) { 
+        var data = this.nodeInstance.rpf.wrapper.fRunUpdater( vni.inputStates );
+        if ( data ) {
+            var outputState = stateFactory( vnid, data );
```
dbooth-boston, line 60: This should just set the data property of the existing output state object, instead of creating a new state object.

Also, it should generate and set a new LM in the state.

Also, as future TODO, it should set the previousLms from the inputStates.

src/framework.js
```diff
@@ -0,0 +1,194 @@
...
+            vni.outputState = outputState;
+
+            // TODO: Discuss whether this should be in the access layer instead of here
+            this.nodeInstance.outPorts.output.send( outputState )
+            this.nodeInstance.outPorts.output.disconnect();
+
+            // Clean up input states
+            vni.inputStates.deleteAll();
```
dbooth-boston, line 68: The inputStates should never be deleted unless the VNI is destroyed.

src/framework.js
```diff
@@ -0,0 +1,194 @@
...
+ * Check the input to see if we have all data required to run the updater or not. 
+ * This is simply a default updater policy - we may have other policies in the future.
+ * 
+ * @param vni virtual node instance whose input states are to be checked.
+ *
+ * @return true if we have received data on all required input port edges
+ */
+function shouldRunUpdater( vni ) { 
+
+    var requiredInPorts = requiredInputPorts.call( vni.node );
```
dbooth-boston, line 85: Actually it needs to check all _attached_ ports -- not required ports.   The 'required' property refers to a config-time requirement that something be attached to it -- either an IIP or an edge attachment.  But at runtime, we need to know if we have data from all of the attached ports, whether they are required or optional.

Also, one other issue we'll have to deal with: If a port is not required and is not multi and has nothing attached, there will still be a corresponding updater parameter, which should receive an undefined value.  The easiest way to deal with this case might be to create a fake (default) IIP to pass in, with undefined data.  This could be set when the inputStates object is created, since it can be determined statically.

dbooth-boston:
I like this overall, though there are a few things that need to be fixed.  I have not finished reviewing all of it yet.

test/merge-patient-lab-iips-mocha.js
```diff
@@ -26,38 +26,31 @@ describe('merge-patient-lab-iips', function() {
             });
         });
 
-        it("should have a vni function", function() {
+        it("should have a vnis object", function() {
```
catch-point, line 29: Can you pull the changes to the mocha out into a separate PR?

src/noflo-component-factory.js
```diff
@@ -22,22 +22,40 @@ var noflo = require('noflo');
...
         registerPorts(nodeInstance.outPorts, nodeDef.outPorts);
         registerPorts(nodeInstance.inPorts, nodeDef.inPorts);
         registerListeners(nodeInstance, nodeDef);
+
         nodeInstance.description = nodeDef.description;
         nodeInstance.setIcon(nodeDef.icon);
+
+        // If a nodeExt object was defined, extend the node instance with it
+        if ( _.isObject( nodeExt ) ) { 
+            nodeInstance = _.extend( nodeInstance, nodeExt );
```
catch-point, line 56: I think this would be better left to be done by the caller

src/input-states.js
```diff
@@ -0,0 +1,256 @@
+/**
+ * input-states.js
```
catch-point, line 2: How is this different from the input-states PR?

src/framework.js
```diff
@@ -0,0 +1,194 @@
...
+ * TODO: Move this to the noflo access layer
+ * 
+ * @param port a noflo port object
+ * @param inputCount the number of inputs received for this port so far.
+ *
+ * @return true if the inputCount matches the number of sockets for the port.
+ *         There will be one socket for each connected input edge.  
+ */
+function hasAllInputs( port, inputCount ) {
+    return (port.sockets.length === inputCount);
```
catch-point, line 116: Use port.listAttached().length here

dbooth-boston:
I think I finished reviewing this.  LMK when it is updates, so that I can re-review.

PR #45, opened 2016-02-22
input-states module

glenna-m:
An input-states implementation (my latest)

### DIFF
--- a/src/input-states.js
+++ b/src/input-states.js
@@ -1,0 +1,256 @@
+/**
+ * input-states.js
+ * 
+ * Manages RDF pipeline input states for a VNI
+ * TODO: Add optimization for input states as a sorted array
+ */
+
+var _ = require('underscore');
+
+var InputType = {
+    IIP: 1,
+    PACKET: 2
+};
+var ALL_PORTINFO_KEYS = [ 'inportName', 'sourceNodeName', 'sourcePortName' ];
+
+var inputStates = [];
+
+/**
+ * Get the index to the state in the input states array if it exists already
+ * 
+ * @param portInfo an object that describes the port(s) associated with the data
+ *
+ * @return an integer index into the inputStates array if a state is found or -1 if no state was found.
+ */
+function getStateIndex( portInfo ) { 
+
+    var portSourceNodeName = portInfo.sourceNodeName || '';
+    var portSourcePortName = portInfo.sourcePortName || '';
+
+    return _.findIndex( inputStates, function(inputState) { 
+        
+        var stateSourceNodeName = inputState.sourceNodeName || '';
+        var stateSourcePortName = inputState.sourcePortName || '';
+
+        return ( portInfo.inportName === inputState.inportName &&
+                 portSourceNodeName === stateSourceNodeName &&
+                 portSourcePortName === stateSourcePortName );
+    });
+}
+
+
+function insertAtIndex( portInfo ) { 
+
+    var portSourceNodeName = portInfo.sourceNodeName || '';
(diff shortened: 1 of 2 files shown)


src/input-states.js
```diff
@@ -0,0 +1,256 @@
...
+        var stateSourcePortName = inputState.sourcePortName || '';
+
+        return ( portInfo.inportName === inputState.inportName &&
+                 portSourceNodeName === stateSourceNodeName &&
+                 portSourcePortName === stateSourcePortName );
+    });
+}
+
+
+function insertAtIndex( portInfo ) {
```
dbooth-boston, line 42: Suggest renaming this to insertionIndex, so that it does not sound like it is performing the actual insertion.

src/input-states.js
```diff
@@ -0,0 +1,256 @@
...
+        var inputState = inputStates[i];
+        var stateSourceNodeName = inputState.sourceNodeName || '';
+        var stateSourcePortName = inputState.sourcePortName || '';
+
+        if ( portInfo.inportName <= inputState.inportName &&
+             portSourceNodeName <= stateSourceNodeName &&
+             portSourcePortName <= stateSourcePortName ) {
+             
+             return i;
+        }
```
dbooth-boston, line 58: I think it would be good to insert them in sorted order, but I don't think the above 'if' statement does that.  I suggest something like the following instead of the above 'if':

<pre>
// Insert in sorted order
if (portInfo.inportName < inputState.inportName) { return i; }
if (portInfo.inportName > inputState.inportName) { continue; }
// portInfo.inportName === inputState.inportName
if (portSourceNodeName < stateSourceNodeName) { return i; }
if (portSourceNodeName > stateSourceNodeName) { continue; }
// portSourceNodeName === stateSourceNodeName
if (portSourcePortName < stateSourcePortName) { return i; }
if (portSourcePortName > stateSourcePortName) { continue; }
// portSourcePortName === stateSourcePortName
break;
</pre>

src/input-states.js
```diff
@@ -0,0 +1,256 @@
...
+        if ( index !== -1 ) { 
+            inputStates.splice(index, 1);
+        }
+ 
+        return this;
+    },
+
+    /** 
+     * Get all input data for the specified port.  If there are multiple inputs on this port,
+     * an array of the input data will be returned.
```
dbooth-boston, line 162: This should return an array iff the port is multi, so that the updater function predictably receives either an array if the port is multi, regardless of how many things are actually attached.

src/input-states.js
```diff
@@ -0,0 +1,256 @@
...
+     * @param portName name of port for which all state data should be retrieved
+     *
+     * @return all port data associated with the portName
+     */
+    getData: function( portName ) { 
+
+        var states = _.pluck( _.where( inputStates, { inportName: portName } ), 'state' );
+
+        if ( _.isEmpty( states ) ) { 
+            return {};
```
dbooth-boston, line 173: This should return undefined if the port is not multi; otherwise return [].

src/input-states.js
```diff
@@ -0,0 +1,256 @@
...
+        } 
+
+        if ( states.length === 1 ) {
```
dbooth-boston, line 176: This should check for non-multi instead of states.length .

Somewhere there should also be error checking: (a) to ensure that required ports are attached; and (b) to detect multiple attachments to a non-multi port.   It probably should not be done here, so I've created issue 24 for it: https://github.com/rdf-pipeline/noflo-rdf-pipeline/issues/24

dbooth-boston:
LGTM, subject to comments above.

PR #46, opened 2016-02-23
Input states function

catch-point:
This pull request changes the input-states module to be a single function and uses arguments.length and type to determine behaviour (i.e. getter vs setter)

### DIFF
--- a/src/input-states.js
+++ b/src/input-states.js
@@ -5,7 +5,7 @@ var _ = require('underscore');
 /**
  * Getter and Setter for InPort states.
  *
- * @this is not used
+ * @this is not used, but may be returned
  * @param node a node facade
  * @param vnid an identifier that distinguishes the set of VNI states
  * @param portName name of the input port whose state is being recorded
@@ -23,20 +23,23 @@ var _ = require('underscore');
  *  inputState(node, vnid, portName, socketId, state) : this
  */ 
 module.exports = function(node, vnid, portName, socketId, state) {
-    if (!node.inPorts || !node.outPorts)
-        throw Error("This isn't a Component node");
-    else if (!_.isString(vnid))
+    if (!node.inPorts || !node.outPorts) {
+        throw Error("This isn't a component facade node");
+    } else if (!_.isString(vnid)) {
         throw Error("Invalid vnid: " + vnid);
-    else if (_.isString(portName) && !node.inPorts[portName])
+    } else if (_.isString(portName) && !node.inPorts[portName]) {
         throw Error("Invalid portName: " + portName);
-    else if (arguments.length > 4)
-        return setPortStateBySocketId(node, vnid, portName, socketId, state);
-    else if (_.isString(portName))
+    } else if (arguments.length > 4) {
+        setPortStateBySocketId(node, vnid, portName, socketId, state);
+        return this;
+    } else if (_.isString(portName)) {
         return getPortStateBySocketId(node, vnid, portName, socketId);
-    else if (_.isObject(arguments[2]))
-        return setAllPortStates(node, vnid, arguments[2]);
-    else
+    } else if (_.isObject(arguments[2])) {
+        setAllPortStates(node, vnid, arguments[2]);
+        return this;
+    } else {
         return getAllPortStates(node, vnid);
+    }
 }
 
 /**
@@ -124,7 +127,6 @@ function setAllPortStates(node, vnid, portStates) {
     _.each(portStates, function(state, portName) {
         setPortState(node, vnid, portName, state);
     });
-    return node;
 }
 
 /**
@@ -147,14 +149,13 @@ function setPortState(node, vnid, portName, state) {
             }
         }
     } else if (port.isMulti()) {
-        return setPortStateArray(node, vnid, portName, state);
+        setPortStateArray(node, vnid, portName, state);
     } else {
         if (!vniStateExists(node, vnid, portName)) {
             node.vnis[vnid].inputStates = node.vnis[vnid].inputStates || {};
         }
         node.vnis[vnid].inputStates[portName] = state;
     }
-    return node;
 }
 
 /**
@@ -172,7 +173,6 @@ function setPortStateArray(node, vnid, portName, stateArray) {
     port.listAttached().forEach(function(socketId, i) {
         setPortStateBySocketId(node, vnid, portName, socketId, stateArray[i]);
     }, []);
-    return node;
 }
 
 /**
@@ -186,11 +186,11 @@ function setPortStateArray(node, vnid, portName, stateArray) {
  * @param state the new state of this port socket
  */
 function setPortStateBySocketId(node, vnid, portName, socketId, state) {
-    if (socketId == null)
-        return setPortState(node, vnid, portName, state);
-    else if (!node.inPorts[portName].isMulti())
+    if (socketId == null) {
+        setPortState(node, vnid, portName, state);
+    } else if (!node.inPorts[portName].isMulti()) {
         throw Error("This port is not addressable/multi: " + portName);
-    else if (_.isUndefined(state)) {
+    } else if (_.isUndefined(state)) {
         if (vniStateExists(node, vnid, portName)) {
             node.vnis[vnid].inputStates[portName][socketId] = state;
             if (!_.compact(node.vnis[vnid].inputStates[portName]).length) {
@@ -207,7 +207,6 @@ function setPortStateBySocketId(node, vnid, portName, socketId, state) {
         }
         node.vnis[vnid].inputStates[portName][socketId] = state;
     }
-    return node;
 }
 
 /**
--- a/src/input-states.js
+++ b/src/input-states.js
@@ -96,6 +96,9 @@ function getPortStateArray(node, vnid, portName) {
  * Returns the state of this port for the given socketId. This can be used to
  * look up the last state when checking to see if the state has changed.
  * Note that the socketId is not neccessarily the same as the array index
+ * If a socket is detached, for example, its socketId is not reused, but the
+ * array index used by listAttached() is. listAttached() maps *attached* sockets
+ * to socketIds.
  * @param node a node facade
  * @param vnid an identifier that distinguishes the set of VNI states
  * @param portName the name of the addressable/multi port for this node
@@ -172,7 +175,7 @@ function setPortStateArray(node, vnid, portName, stateArray) {
         throw Error("This port is not addressable/multi: " + portName);
     port.listAttached().forEach(function(socketId, i) {
         setPortStateBySocketId(node, vnid, portName, socketId, stateArray[i]);
-    }, []);
+    });
 }
 
 /**
--- a/src/input-states.js
+++ b/src/input-states.js
@@ -5,6 +5,7 @@ var _ = require('underscore');
 /**
  * Getter and Setter for InPort states.
  *
+ * @this is not used
  * @param node a node facade
  * @param vnid an identifier that distinguishes the set of VNI states
  * @param portName name of the input port whose state is being recorded
@@ -15,10 +16,10 @@ var _ = require('underscore');
  *         setter returns the current context
  *
  * Usage to retrieve state:
- *  inputState(node, vnid) : {portName: currentState, addressablePortName: [currentState]}
+ *  inputState(node, vnid) : {portName: currentState, multiPortName: [currentState]}
  *  inputState(node, vnid, portName, socketId) : currentState
  * Usage to store state:
- *  inputState(node, vnid, {portName: currentState, addressablePortName: [currentState]}) : this
+ *  inputState(node, vnid, {portName: currentState, multiPortName: [currentState]}) : this
  *  inputState(node, vnid, portName, socketId, state) : this
  */ 
 module.exports = function(node, vnid, portName, socketId, state) {
--- a/src/input-states.js
+++ b/src/input-states.js
@@ -62,7 +62,7 @@ function getAllPortStates(node, vnid) {
 function getPortState(node, vnid, portName) {
     var port = node.inPorts[portName];
     if (port.isMulti()) return getPortStateArray(node, vnid, portName);
-    else if (ensureVniStateExists(node, vnid, portName))
+    else if (vniStateExists(node, vnid, portName))
         return node.vnis[vnid].inputStates[portName];
     // By default states that come in on vnid '' apply to all VNIs
     else if (vnid) return getPortState(node, '', portName);
@@ -99,7 +99,7 @@ function getPortStateBySocketId(node, vnid, portName, socketId) {
         return getPortState(node, vnid, portName);
     else if (!isMultiPort(node.inPorts[portName]))
         throw Error("This port is not addressable/multi: " + portName);
-    else if (ensureVniStateExists(node, vnid, portName) &&
+    else if (vniStateExists(node, vnid, portName) &&
             node.vnis[vnid].inputStates[portName][socketId])
         return node.vnis[vnid].inputStates[portName][socketId];
     // By default states that come in on vnid '' apply to all VNIs
@@ -133,21 +133,16 @@ function setAllPortStates(node, vnid, portStates) {
 function setPortState(node, vnid, portName, state) {
     var port = node.inPorts[portName];
     if (_.isUndefined(state)) {
-        if (ensureVniStateExists(node, vnid, portName)) {
+        if (vniStateExists(node, vnid, portName)) {
             delete node.vnis[vnid].inputStates[portName];
             if (_.isEmpty(node.vnis[vnid].inputStates)) {
                 delete node.vnis[vnid].inputStates;
-                if (_.isEmpty(node.vnis[vnid])) {
-                    delete node.vnis[vnid];
-                }
             }
         }
     } else if (port.isMulti()) {
         return setPortStateArray(node, vnid, portName, state);
     } else {
-        if (!ensureVniStateExists(node, vnid, portName)) {
-            node.vnis = node.vnis || {};
-            node.vnis[vnid] = node.vnis[vnid] || {};
+        if (!vniStateExists(node, vnid, portName)) {
             node.vnis[vnid].inputStates = node.vnis[vnid].inputStates || {};
         }
         node.vnis[vnid].inputStates[portName] = state;
@@ -188,22 +183,17 @@ function setPortStateBySocketId(node, vnid, portName, socketId, state) {
     else if (!isMultiPort(node.inPorts[portName]))
         throw Error("This port is not addressable/multi: " + portName);
     else if (_.isUndefined(state)) {
-        if (ensureVniStateExists(node, vnid, portName)) {
+        if (vniStateExists(node, vnid, portName)) {
             node.vnis[vnid].inputStates[portName][socketId] = state;
             if (!_.compact(node.vnis[vnid].inputStates[portName]).length) {
                 delete node.vnis[vnid].inputStates[portName];
                 if (_.isEmpty(node.vnis[vnid].inputStates)) {
                     delete node.vnis[vnid].inputStates;
-                    if (_.isEmpty(node.vnis[vnid])) {
-                        delete node.vnis[vnid];
-                    }
                 }
             }
         }
     } else {
-        if (!ensureVniStateExists(node, vnid, portName)) {
-            node.vnis = node.vnis || {};
-            node.vnis[vnid] = node.vnis[vnid] || {};
+        if (!vniStateExists(node, vnid, portName)) {
             node.vnis[vnid].inputStates = node.vnis[vnid].inputStates || {};
             node.vnis[vnid].inputStates[portName] = node.vnis[vnid].inputStates[portName] || [];
         }
@@ -238,7 +228,7 @@ function isPort(port) {
  * @param vnid an identifier that distinguishes the set of VNI states
  * @param portName name of the port on this node
  */
-function ensureVniStateExists(node, vnid, portName) {
+function vniStateExists(node, vnid, portName) {
     return node.vnis && node.vnis[vnid] &&
         node.vnis[vnid].inputStates &&
         node.vnis[vnid].inputStates[portName];
--- a/src/input-states.js
+++ b/src/input-states.js
@@ -45,8 +45,7 @@ module.exports = function(node, vnid, portName, socketId, state) {
  * @param vnid an identifier that distinguishes the set of VNI states
  */
 function getAllPortStates(node, vnid) {
-    var inPorts = _.pick(node.inPorts, isPort);
-    return _.mapObject(inPorts, function(port, portName){
+    return _.mapObject(node.inPorts, function(port, portName){
         return getPortState(node, vnid, portName);
     });
 }
@@ -79,7 +78,7 @@ function getPortState(node, vnid, portName) {
  */
 function getPortStateArray(node, vnid, portName) {
     var port = node.inPorts[portName];
-    if (!isMultiPort(port)) throw Error("This port is not addressable/multi: " + portName);
+    if (!port.isMulti()) throw Error("This port is not addressable/multi: " + portName);
     else return port.listAttached().map(function(socketId){
         return getPortStateBySocketId(node, vnid, portName, socketId);
     });
@@ -97,7 +96,7 @@ function getPortStateArray(node, vnid, portName) {
 function getPortStateBySocketId(node, vnid, portName, socketId) {
     if (socketId == null)
         return getPortState(node, vnid, portName);
-    else if (!isMultiPort(node.inPorts[portName]))
+    else if (!node.inPorts[portName].isMulti())
         throw Error("This port is not addressable/multi: " + portName);
     else if (vniStateExists(node, vnid, portName) &&
             node.vnis[vnid].inputStates[portName][socketId])
@@ -160,7 +159,7 @@ function setPortState(node, vnid, portName, state) {
  */
 function setPortStateArray(node, vnid, portName, stateArray) {
     var port = node.inPorts[portName];
-    if (!isMultiPort(port))
+    if (!port.isMulti())
         throw Error("This port is not addressable/multi: " + portName);
     port.listAttached().forEach(function(socketId, i) {
         setPortStateBySocketId(node, vnid, portName, socketId, stateArray[i]);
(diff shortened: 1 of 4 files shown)


src/input-states.js
```diff
@@ -0,0 +1,289 @@
...
+ * Returns the state of the port or an array of states, if the port is
+ * addressable. When all the port states are not need this is a convient way to
+ * retrieve just the port state that is needed.
+ * @param component a Component or a component facade
+ * @param vnid an identifier that distinguishes the set of states
+ * @param portName the InPort name on this component 
+ */
+function getPortState(component, vnid, portName) {
+    var port = component.inPorts[portName];
+    if (port.isAddressable()) return getPortStateArray(component, vnid, portName);
```
dbooth-boston: Let's use the term 'multi' instead of 'addressable' in Framework code.  The only place 'addressable' should be used would be in the noflo access layer.

src/input-states.js
```diff
@@ -0,0 +1,289 @@
...
+ * all the port states are not need this is a convient way to
+ * retrieve just the port state that is needed.
+ * @param component a Component or a component facade
+ * @param vnid an identifier that distinguishes the set of states
+ * @param portName the InPort name on this component 
+ */
+function getPortStateArray(component, vnid, portName) {
+    var port = component.inPorts[portName];
+    var state = component.vnis && component.vnis[vnid] && component.vnis[vnid][portName];
+    if (!isAddressablePort(port)) throw Error("This port is not addressable: " + portName);
```
dbooth-boston: s/addressable/multi/g

src/input-states.js
```diff
@@ -0,0 +1,289 @@
...
+ * retrieve just the port state that is needed.
+ * @param component a component facade
+ * @param vnid an identifier that distinguishes the set of states
+ * @param portName the InPort name on this component 
+ */
+function getPortState(component, vnid, portName) {
+    var port = component.inPorts[portName];
+    if (port.isMulti()) return getPortStateArray(component, vnid, portName);
+    else if (component.vnis && component.vnis[vnid] && component.vnis[vnid][portName])
+        return component.vnis[vnid][portName];
```
dbooth-boston: This does not look like the VNI data structure that i am expecting.  What is component.vnis[vnid][portName]?

src/input-states.js
```diff
@@ -0,0 +1,289 @@
...
+    else if (vnid) return getPortState(component, '', portName);
```
dbooth-boston: This logic needs to be at the per-attachment level, not at the per-port level, because a port might have both an IIP attached and an edge attachment to another node.  Also, defaulting to vnid of '' needs to happen _only_ for an IIP (so you need to explicitly see if the attachment is an IIP), because a node like 'split' may have states both for vnid '' and for other vnids.

catch-point: I not sure I follow all of this, but before we can distinguish between IIP/upstream the port needs to be addressable and this function does not handle addressable ports: see getPortStateIndex function for that.

src/input-states.js
catch-point: For addressable ports it is an array of states, or non-addressable ports it is the most recent state state.

src/noflo-component-access.js
```diff
@@ -34,7 +34,15 @@ function facadePort(nodeInstance, port, name) {
     return _.extend({
         name: name,
         nodeInstance: nodeInstance,
-        isMulti: port.isAddressable.bind(port)
+        isMulti: port.isAddressable.bind(port),
+        getComponentIdOn: function(socketIndex) {
```
dbooth-boston: s/getComponentIdOn/getComponentNameOn/

src/input-states.js
dbooth-boston: I would think it should be component.vnis[vnid].inputStates[portName], to avoid collisions between portNames and other properties we may want to hang on the VNI.

src/input-states.js
dbooth-boston: Oh, right, now I see that this line only applies to a non-multi port.  

Maybe add a comment above this line, like:
// By default states that come in on vnid '' apply to all VNIs

src/input-states.js
```diff
@@ -0,0 +1,289 @@
...
+/**
+ * Returns the state of this port for the given socketIndex. This can be used to
+ * look up the last state when checking to see if the state has changed.
+ * Note that the socketIndex is not neccessarily the same as the array index
+ * @param component a component facade
+ * @param vnid an identifier that distinguishes the set of states
+ * @param portName the name of the addressable/multi port for this component
+ * @param socketIndex the socketIndex (or socketId) provided to a port data event
+ */
+function getPortStateIndex(component, vnid, portName, socketIndex) {
```
dbooth-boston: Suggest renaming getPortStateIndex -> getPortStateBySocketId

src/input-states.js
```diff
@@ -0,0 +1,289 @@
...
+    if (!isMultiPort(component.inPorts[portName]))
+        throw Error("This port is not addressable/multi: " + portName);
+    component.vnis = component.vnis || {};
+    component.vnis[vnid] = component.vnis[vnid] || {};
```
dbooth-boston: I think it would be better to use a generic ensureVniExists function here, since this is likely to be a common need:

<pre>
var vni = ensureVniExists(component, vnid);
</pre>

src/input-states.js
```diff
@@ -0,0 +1,289 @@
...
+    component.vnis[vnid][portName] = component.vnis[vnid][portName] || [];
+    if (component.vnis[vnid][portName].length)
+        return component.vnis[vnid][portName][socketIndex];
+    else if (vnid) return getPortStateIndex(component, '', portName, socketIndex);
```
dbooth-boston: Maybe add a comment above this line, like:
// By default states that come in on vnid '' apply to all VNIs

src/input-states.js
```diff
@@ -0,0 +1,289 @@
...
+    return component;
+}
+
+/**
+ * Changes the state of a port. When a new state is received on a port this
+ * should be called.
+ * @param component a component facade
+ * @param vnid an identifier that distinguishes the set of states
+ * @param portName name of the port on this component
+ * @param state the new state to assign this port to
```
dbooth-boston: @param state the new state (or array of states if the port is multi/addressable) to assign to this port

src/input-states.js
```diff
@@ -0,0 +1,289 @@
...
+            delete component.vnis[vnid][portName];
+            if (_.isEmpty(component.vnis[vnid])) {
+                delete component.vnis[vnid];
+            }
+        }
+    } else if (port.isMulti()) {
+        return setPortStateArray(component, vnid, portName, state);
+    } else {
+        component.vnis = component.vnis || {};
+        component.vnis[vnid] = component.vnis[vnid] || {};
```
dbooth-boston: Again,  I think it would be better to use a common ensureVniExists function.

src/input-states.js
```diff
@@ -0,0 +1,289 @@
...
+        throw Error("This port is not addressable/multi: " + portName);
+    var socketIndex = findSocketIndex(component, portName, sourceNode, sourcePort);
+    return setPortStateIndex(component, vnid, portName, socketIndex, state);
+}
+
+/**
+ * Changes the state of a addressable/multi port socket. When a new state is received
+ * on a port this should be called.
+ * @param component a component facade
+ * @param vnid an identifier that distinguishes the set of states
```
dbooth-boston: s/states/VNIs/

src/input-states.js
```diff
@@ -0,0 +1,289 @@
...
+ * @param portName name of the port on this component
+ * @param socketIndex the socketIndex (or socketId) pass to the data event handler
```
dbooth-boston: Since there seem to be two indexes involved (as mentioned on line 132), maybe it would be better to consistently use the term socketId instead of socketIndex to distinguish it from the other array index.

src/input-states.js
```diff
@@ -0,0 +1,289 @@
...
+ * @param state the new state of this port socket
+ */
+function setPortStateIndex(component, vnid, portName, socketIndex, state) {
```
dbooth-boston: Suggest renaming setPortStateIndex -> setPortStateBySocketId

src/input-states.js
```diff
@@ -0,0 +1,289 @@
...
+    if (!isMultiPort(component.inPorts[portName]))
+        throw Error("This port is not addressable/multi: " + portName);
+    component.vnis = component.vnis || {};
+    component.vnis[vnid] = component.vnis[vnid] || {};
```
dbooth-boston: ensureVniExists

src/input-states.js
```diff
@@ -0,0 +1,289 @@
...
+    component.vnis[vnid][portName] = component.vnis[vnid][portName] || [];
+    component.vnis[vnid][portName][socketIndex] = state;
+    if (_.isUndefined(state) && !_.compact(component.vnis[vnid][portName]).length) {
+        delete component.vnis[vnid][portName];
```
dbooth-boston: Why are these being deleted when they're empty?

src/input-states.js
```diff
@@ -0,0 +1,289 @@
...
+ * Given the source Component node name and source OutPort name, returns the
+ * socketIndex (aka socketId) of the addressable/multi InPort with the name portName.
+ * This maps the sourceNode/sourcePort to socketIndex to allow either form to be
+ * used in a seamless way.
+ * @param component a component facade
+ * @param portName the addressable/multi InPort of this component
+ * @param sourceNode the node name of the upstream component
+ * @param sourcePort the OutPort name of the upstream component's port
+ */
+function findSocketIndex(component, portName, sourceNode, sourcePort) {
```
dbooth-boston: Maybe s/findSocketIndex/findSocketId/

src/input-states.js
```diff
@@ -0,0 +1,289 @@
...
+ */
+function isPort(port) {
+    return _.isFunction(port.isMulti);
+}
+
+/**
+ * Given the source Component node name and source OutPort name, returns the
+ * socketIndex (aka socketId) of the addressable/multi InPort with the name portName.
+ * This maps the sourceNode/sourcePort to socketIndex to allow either form to be
+ * used in a seamless way.
```
dbooth-boston: Nice

src/input-states.js
```diff
@@ -0,0 +1,289 @@
...
+ * @param component a component facade
+ * @param portName the addressable/multi InPort of this component
+ * @param sourceNode the node name of the upstream component
+ * @param sourcePort the OutPort name of the upstream component's port
+ */
+function findSocketIndex(component, portName, sourceNode, sourcePort) {
+    var port = component.inPorts[portName];
+    var socketIndex = _.find(port.listAttached(), function(socketIndex) {
+        return port.getComponentPortNameOn(socketIndex) == sourcePort &&
+            port.getComponentIdOn(socketIndex) == sourceNode;
```
dbooth-boston: Suggest renaming getComponentIdOn -> getComponentNameOn

test/input-states-mocha.js
```diff
@@ -0,0 +1,341 @@
...
+describe('input-states', function() {
+    var component, network;
+    var oninput, onoutput;
+    beforeEach(function(){
+        var getComponent = function() {
+            var factory = componentFactory({
+                inPorts:{
+                    input:{
+                        ondata: function(payload) {
+                            oninput(payload);
```
dbooth-boston: I would have thought that our componentFactory would already set the ondata function, so it would not have to be set explicitly here.  Maybe that is not implemented yet?

test/input-states-mocha.js
```diff
@@ -0,0 +1,341 @@
+/**
+ * File: input-state-mocha.js
+ * Unit tests for the input-state APIs defined in src/input-state.js
```
dbooth-boston: Does this test for the following cases?

 - An optional port that is not attached to anything should return an undefined state.

 - A multi port with nothing attached should return an empty list of states.

 - Ensure that required ports are attached and check for too many attachments to non-multi ports ( https://github.com/rdf-pipeline/noflo-rdf-pipeline/issues/24 ).  These checks do not necessarily have to be done in this module though.

 - Getting the states from a multi port with both an IIP and an edge attached, with a vnid of '' coming on the edge attachment.

 - Getting the states from a multi port with both an IIP and an edge attached, with a non-empty-string vnid coming on the edge attachment.

dbooth-boston:
Overall this code is very nicely written, and considerably easier to read than some of your previous code. (Thanks for that!)  I am still finding the test cases a bit hard to read though.  There is no need to change them now, but if you can think of ways in the future to make your test cases easier for me to follow I would appreciate it.  Thanks!

dbooth-boston:
However, I should add that, although we discussed the idea of using a single function for all this functionality, in hind sight I think it was a mistake, because: (a) it is harder to know what a call of inputState(...) is really doing, because the reader needs to count the parameters to figure out what function is _really_ being called; (b) we're stuck when we discover that we want to add a new function to the module; and (c) it adds unnecessary runtime cost doing the dispatch (even if it's small).  

For getter/setter pairs I think it makes sense to use a single function, but otherwise it would be clearer -- and faster at runtime -- to expose the different functions directly.

src/noflo-component-access.js
catch-point: Noflo uses the term componentName to identify the factory of the nodeInstance. I think it would be too confusing to adopt the same term to mean something slightly different half the time.

src/input-states.js
catch-point: The if condition checks that they are empty. When setting a state to undefined the vnis hash is compacted to avoid memory leaks.

src/input-states.js
```diff
@@ -0,0 +1,236 @@
...
+ *
+ * @return getter returns the current input port state state.
+ *         setter returns the current context
+ *
+ * Usage to retrieve state:
+ *  inputState(node, vnid) : {portName: currentState, multiPortName: [currentState]}
+ *  inputState(node, vnid, portName, socketId) : currentState
+ * Usage to store state:
+ *  inputState(node, vnid, {portName: currentState, multiPortName: [currentState]}) : this
+ *  inputState(node, vnid, portName, socketId, state) : this
```
dbooth-boston: I'm still not convinced that there is advantage here in hiding these 4 functions behind one inputState function, but if both James and Glenna prefer it that way then it's ok with me.

src/input-states.js
```diff
@@ -0,0 +1,236 @@
...
+/**
+ * Returns a hash of port names to port states or array of states if the port is
+ * addressable/multi. This provides the caller with all portNames, distingishes between
+ * addressable/multi and non-addressable/multi ports, and provides the caller with all their
+ * states that will be useful when calling an updater.
+ * @param node a node facade
+ * @param vnid an identifier that distinguishes the set of VNI states
+ */
+function getAllPortStates(node, vnid) {
+    var inPorts = _.pick(node.inPorts, isPort);
```
dbooth-boston: Is _.pick really needed here?  Why would there be anything but ports in inPorts?

src/input-states.js
```diff
@@ -0,0 +1,236 @@
...
+    return _.mapObject(inPorts, function(port, portName){
+        return getPortState(node, vnid, portName);
```
dbooth-boston, line 55: nice

src/input-states.js
```diff
@@ -0,0 +1,236 @@
...
+
+/**
+ * Returns the state of the port or an array of states, if the port is
+ * addressable/multi. When all the port states are not need this is a convient way to
+ * retrieve just the port state that is needed.
+ * @param node a node facade
+ * @param vnid an identifier that distinguishes the set of VNI states
+ * @param portName the InPort name on this node 
+ */
+function getPortState(node, vnid, portName) {
```
dbooth-boston: nice

src/input-states.js
```diff
@@ -0,0 +1,236 @@
...
+
+/**
+ * Returns an array of the port states, one item for each attached socket. When
+ * all the port states are not need this is a convient way to
+ * retrieve just the port state that is needed.
+ * @param node a node facade
+ * @param vnid an identifier that distinguishes the set of VNI states
+ * @param portName the InPort name on this node 
+ */
+function getPortStateArray(node, vnid, portName) {
```
dbooth-boston: nice

src/input-states.js
```diff
@@ -0,0 +1,236 @@
...
+ * Returns the state of this port for the given socketId. This can be used to
+ * look up the last state when checking to see if the state has changed.
+ * Note that the socketId is not neccessarily the same as the array index
+ * @param node a node facade
+ * @param vnid an identifier that distinguishes the set of VNI states
+ * @param portName the name of the addressable/multi port for this node
+ * @param socketId the socketId provided to a port data event
+ */
+function getPortStateBySocketId(node, vnid, portName, socketId) {
+    if (socketId == null)
```
dbooth-boston: JS noobie question: Is (socketId == null) the same as checking _.isUndefined(socketId)?

src/input-states.js
```diff
@@ -0,0 +1,236 @@
...
+        return getPortState(node, vnid, portName);
```
dbooth-boston: okay

src/input-states.js
```diff
@@ -0,0 +1,236 @@
...
+    else if (!isMultiPort(node.inPorts[portName]))
+        throw Error("This port is not addressable/multi: " + portName);
+    else if (vniStateExists(node, vnid, portName) &&
+            node.vnis[vnid].inputStates[portName][socketId])
+        return node.vnis[vnid].inputStates[portName][socketId];
+    // By default states that come in on vnid '' apply to all VNIs
+    else if (vnid) return getPortStateBySocketId(node, '', portName, socketId);
```
dbooth-boston: good

src/input-states.js
```diff
@@ -0,0 +1,236 @@
...
+}
+
+/**
+ * Changes the state of each given port. If a port name is missing it is unmodified.
+ * This can be run in a test environment to reset the state of the ports.
+ * @param node a node facade
+ * @param vnid an identifier that distinguishes the set of VNI states
+ * @param portStates hash of portName to state
+ */
+function setAllPortStates(node, vnid, portStates) {
```
dbooth-boston: ok

src/input-states.js
catch-point: noflo has other things on inPorts, but our facade strips them out already. I thought it didn't hurt leaving this in.

src/input-states.js
jamesrdf: socketId == null true if the socketId is null or undefined

src/input-states.js
```diff
@@ -0,0 +1,236 @@
...
+}
+
+/**
+ * Checks that the given object is a port or a port facade. Noflo does not provide
+ * a list of available port in a node and this allows the caller to remove
+ * other properties and functions that are on the inPorts objects.
+ * @param port a port facade.
+ */
+function isPort(port) {
+    return _.isFunction(port.isMulti);
```
dbooth-boston: That's an odd way to check whether the object is a port.  Maybe this should be moved to the access layer, since it relies on internal knowledge of how we're doing the port facade.

src/input-states.js
```diff
@@ -0,0 +1,236 @@
...
+ * Changes the state of a addressable/multi port socket. When a new state is received
+ * on a port this should be called.
+ * setPortState should be called instead if the port is non-multi (non-addressable).
+ * @param node a node facade
+ * @param vnid an identifier that distinguishes the set of VNI states
+ * @param portName name of the port on this node
+ * @param socketId the socketId pass to the data event handler
+ * @param state the new state of this port socket
+ */
+function setPortStateBySocketId(node, vnid, portName, socketId, state) {
```
dbooth-boston: I'm okay with this API.

src/input-states.js
```diff
@@ -0,0 +1,236 @@
...
+ * should be called.
+ * setPortStateBySocketId should be used to set the state on a multi/addressable port.
+ * @param node a node facade
+ * @param vnid an identifier that distinguishes the set of VNI states
+ * @param portName name of the port on this node
+ * @param state the new state (or array of states if the port is multi/addressable) to assign to this port
+ */
+function setPortState(node, vnid, portName, state) {
+    var port = node.inPorts[portName];
+    if (_.isUndefined(state)) {
```
dbooth-boston: Let's discuss this.  I don't think the state entry should be deleted from the inputStates hash.  But we do need a way to destroy a state and have the wrapper's destroyState method called -- maybe not in this module though.

As I see it, this function would normally only be used for testing -- not for runtime use -- because at runtime setPortStateBySocketId would be called to set a single state when an event comes in.  In which case, the comment above should be updated to indicate that this function would normally only be used for testing.

src/input-states.js
```diff
@@ -0,0 +1,236 @@
...
+/**
+ * Changes the state of a port. When a new state is received on a port this
+ * should be called.
+ * setPortStateBySocketId should be used to set the state on a multi/addressable port.
+ * @param node a node facade
+ * @param vnid an identifier that distinguishes the set of VNI states
+ * @param portName name of the port on this node
+ * @param state the new state (or array of states if the port is multi/addressable) to assign to this port
+ */
+function setPortState(node, vnid, portName, state) {
```
dbooth-boston: As I see it, this function would normally only be used for testing -- not for runtime use -- because at runtime setPortStateBySocketId would be called to set a single state when an event comes in.  In which case, the comment above should be updated to indicate that this function would normally only be used for testing.

src/input-states.js
jamesrdf: This function is used for non-multi ports, not just for testing

src/input-states.js
```diff
@@ -0,0 +1,215 @@
...
+ *
+ * Usage to retrieve state:
+ *  inputState(node, vnid) : {portName: currentState, multiPortName: [currentState]}
+ *  inputState(node, vnid, portName, socketId) : currentState
+ * Usage to store state:
+ *  inputState(node, vnid, {portName: currentState, multiPortName: [currentState]}) : this
+ *  inputState(node, vnid, portName, socketId, state) : this
+ */ 
+module.exports = function(node, vnid, portName, socketId, state) {
+    if (!node.inPorts || !node.outPorts) throw Error("This isn't a Component node");
```
glenna-m: I thought we had agreed to following the google javascript style guide - https://google.github.io/styleguide/javascriptguide.xml

That style guide recommends braces at the end of the if line for the block, even with a single line.   Are we deliberately changing this?

src/input-states.js
```diff
@@ -0,0 +1,236 @@
...
+ * @param portName name of the port on this node
+ * @param socketId the socketId pass to the data event handler
+ * @param state the new state of this port socket
+ */
+function setPortStateBySocketId(node, vnid, portName, socketId, state) {
+    if (socketId == null)
+        return setPortState(node, vnid, portName, state);
+    else if (!isMultiPort(node.inPorts[portName]))
+        throw Error("This port is not addressable/multi: " + portName);
+    else if (_.isUndefined(state)) {
```
dbooth-boston: Again, same issue as above to discuss.

test/input-states-mocha.js
```diff
@@ -0,0 +1,333 @@
...
+    var node, network;
+    beforeEach(function(){
+        var getComponent = componentFactory({
+            inPorts:{
+                input:{},
+                input1:{},
+                input2:{},
+                input3:{},
+                inputa:{
+                    addressable: true
```
glenna-m: Given we are using isMulti, perhaps one of these should use that setting instead of addressable?

test/input-states-mocha.js
```diff
@@ -0,0 +1,333 @@
...
+
+        var context = node.vni().inputStates({'input': state});
+        context.should.exist;
+
+        var result = node.vni().inputStates('input'); 
+        verifyInputState( result, state );
+    });
+    it('should treat undefined socketIndex the same as without it', function() {
+        var state = { vnid: '001',
+                   data: 'hello world',
```
glenna-m: inconsistent alignment on the data and lm elements here and multiple other places with declaring state values

src/input-states.js
catch-point: Can you quote the style guide? I don't see anything about the use of optional braces.

test/input-states-mocha.js
```diff
@@ -0,0 +1,333 @@
...
+                       data: 'hello world from node one',
+                       lm: createLm() }
+         var two = { vnid: '010',
+                       data: 'hello world from node two',
+                       lm: createLm() }
+
+         var context = node.vni().inputStates({'inputa': [one, two]});
+         context.should.exist;
+
+         node.vni().inputStates('inputa').should.eql([one, two]);
```
glenna-m: This seems a bit odd.  You are setting two states with the same VNID in an array .  Since the vnid is the same for both and on the same port, wouldn't state two overwrite one?   Please add a comment as to what you are doing with this mix.

test/input-states-mocha.js
```diff
@@ -0,0 +1,333 @@
...
+    }); 
+    it('should set and get hash of array of input states', function() {
+         var one = { vnid: '010',
+                       data: 'hello world from node one',
+                       lm: createLm() }
+         var two = { vnid: '010',
+                       data: 'hello world from node two',
+                       lm: createLm() }
+
+         var context = node.vni().inputStates({inputa: [one, two]});
```
glenna-m: I get that this data is supposed to be same vnid from two different nodes, coming into inputa, but I don't see anything set that would ensure input-state treats it that way.    This looks more like a sequence of IIPs with the same vnid where is the set up if this is for multiple source nodes, where is that set up/defined.  Some commenting here might help.

dbooth-boston:
LGTM

src/input-states.js
dbooth-boston: ok

src/input-states.js
```diff
@@ -0,0 +1,222 @@
...
+    if (!port.isMulti()) throw Error("This port is not addressable/multi: " + portName);
+    else return port.listAttached().map(function(socketId){
+        return getPortStateBySocketId(node, vnid, portName, socketId);
+    });
+}
+
+/**
+ * Returns the state of this port for the given socketId. This can be used to
+ * look up the last state when checking to see if the state has changed.
+ * Note that the socketId is not neccessarily the same as the array index
```
dbooth-boston, line 98: Is this comment wrong now?  It looks like socketId is being used directly as the array index on lines 109 and 111.

catch-point, line 98: SocketId is the array index of node.vnis[vnid].inputStates[portName], but not neccessarily the array index of getPortStateArray(node, vnid, portName). The later only includes attached sockets.

dbooth-boston, line 98: ok

dbooth-boston:
LGTM

glenna-m:
LGTM

PR #47, opened 2016-02-23
Noflo component access

catch-point:
Remove EventEmitter functions from facade and remove higher level functions from implementation

### DIFF
--- a/src/noflo-component-access.js
+++ b/src/noflo-component-access.js
@@ -21,7 +21,7 @@ function facadeComponent(component) {
     _.extend(facade, {
         inPorts: _.mapObject(_.pick(component.inPorts, isInPort), facadePort.bind(this, facade)),
         outPorts: _.mapObject(_.pick(component.outPorts, isOutPort), facadePort.bind(this, facade))
-    }, facadeEventEmitter(component));
+    });
     return component._noflo_access_facade = facade;
 }
 
@@ -35,6 +35,8 @@ function facadePort(nodeInstance, port, name) {
         name: name,
         nodeInstance: nodeInstance,
         isMulti: port.isAddressable.bind(port),
+        isRequired: port.isRequired.bind(port),
+        listAttached: port.listAttached.bind(port),
         getComponentIdOn: function(socketIndex) {
             if (port.sockets[socketIndex] && port.sockets[socketIndex].from)
                 return port.sockets[socketIndex].from.process.id;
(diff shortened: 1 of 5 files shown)


dbooth-boston:
LGTM

src/noflo-component-access.js
```diff
@@ -1,4 +1,4 @@
-// noflo-component-access.js
+// noflo-node-access.js
```
glenna-m: This name does not match the file name.

src/noflo-component-access.js
```diff
@@ -6,23 +6,25 @@ var noflo = require('noflo');
...
 
 /**
- * Create a noflo.Component facade from a given component. Access to the component
- * is limited to only select properties and functions. An rpf hash property is
- * also present for arbitrary assignments and is shared for all facades of the
- * same Component instance.
- * This facade mimic the main usage of http://noflojs.org/api/Component.html
+ * Create a noflo.Component facade from a given node. Access to the node
+ * is limited to only select properties and functions.
+ * This facade mimic some of the main usage of http://noflojs.org/api/Component.html
```
glenna-m: Minor nit - should be "This facade **mimics** some of ..."

src/noflo-component-access.js
```diff
@@ -6,23 +6,25 @@ var noflo = require('noflo');
...
-function facadeComponent(component) {
-    if (component._noflo_access_facade)
-        return component._noflo_access_facade;
-    var facade = {};
-    _.extend(facade, {
-        inPorts: _.mapObject(_.pick(component.inPorts, isInPort), facadePort.bind(this, facade)),
-        outPorts: _.mapObject(_.pick(component.outPorts, isOutPort), facadePort.bind(this, facade))
-    }, facadeEventEmitter(component));
-    return component._noflo_access_facade = facade;
+ *  var node = componentAccess(new noflo.Component());
```
glenna-m, line 15: Add @this info to the function comments

test/common-test.js
```diff
@@ -43,9 +43,20 @@ module.exports = {
...
+        var node;
+        var metadata = {
+            facade: function(facade) {
+                node = facade;
+            }
+        };
+        var getComponent = _.isFunction(factory) ? factory : factory.getComponent;
+        var component = getComponent(metadata);
+
+        if (node) {
```
glenna-m: When would we not have a node here?  Can you add a comment on that?

test/common-test.js
```diff
@@ -57,10 +68,11 @@ module.exports = {
             port.name = name;
         });
 
-        return component;
+        return node ? node : component;
```
glenna-m, line 80: This is a little tricky since the function can return a node if it exists or a component if it does not.  Can you add a  comment explaining what's going on with this?

glenna-m:
On the API function facadePort(nodeInstance, port, name)
it would be a little more clear if you made the name parameter "portName"

glenna-m:
The noflo-component-factory has the following lines: 
        if (metadata && _.isFunction(metadata.facade)) {
            metadata.facade.call(node, facade);
        }
I expect this is to allow the framework to add things like the vnis.  Is so?  If some, a comment on it would be good.    This approach is more flexible that what I implemented (which was simply adding whatever metadata attributes were defined), but it is a little more complex.    Why do you have a facade on this?  Do we really need a facade on top of the callback?

src/noflo-component-factory.js
```diff
@@ -28,22 +28,28 @@ var access = require('./noflo-component-access');
...
+        var facade = access(node);
+        registerPorts(node.outPorts, facade.outPorts, nodeDef.outPorts);
+        registerPorts(node.inPorts, facade.inPorts, nodeDef.inPorts);
+        registerListeners(node, facade, nodeDef);
+        node.description = nodeDef.description;
+        node.setIcon(nodeDef.icon);
+        if (_.isFunction(callback)) {
+            callback.call(node, facade);
+        }
+        if (metadata && _.isFunction(metadata.facade)) {
```
glenna-m, line 50: A comment here to explain why we need this would be helpful.

src/noflo-component-factory.js
```diff
@@ -75,27 +81,32 @@ function thenTrigger(fn, event /* arguments to fn and event */) {
  * Register listeners of the set of port definitions with this actual noflo Ports set.
  *
  * Usage:
- *  registerPorts({portName: new Port()}, {
+ *  registerPorts({portName: new Port()}, {portName: new PortFacade()}, {
```
glenna-m, line 85: example needs to be updated to match the current 3 parameter API with facades as an array.

src/noflo-component-factory.js
dbooth-boston, line 50: A comment would be helpful, to say why this is being done

src/noflo-component-factory.js
```diff
@@ -75,27 +81,32 @@ function thenTrigger(fn, event /* arguments to fn and event */) {
...
-        _.each(_.pick(port, isListener), register.bind(ports[name]));
+function registerPorts(ports, facades, portDefs) {
+    _.each(portDefs, function(portDef, name) {
+        registerListeners(ports[name], facades[name], portDef);
     });
 }
 
 /**
  * Registers listers by their event type to this EventEmitter.
+ * @param eventEmitter a noflo.Component/Port
```
dbooth-boston, line 99: s/listers/listeners/

src/noflo-component-factory.js
catch-point, line 85: I don't understand, the comment already includes 3 parameters in the example.

glenna-m:
Should we have a test on the isMulti attribute?  Didn't see anything on that.

test/common-test.js
```diff
@@ -43,9 +43,20 @@ module.exports = {
...
+    createComponent: function(factory) {
 
-        var component = getComponent();
+        var node;
+        var metadata = {
+            facade: function(facade) {
+                node = facade;
+            }
+        };
+        var getComponent = _.isFunction(factory) ? factory : factory.getComponent;
```
dbooth-boston, line 62: Why is factory arriving sometimes a a function and sometimes as an object?  Maybe a comment would help.

src/noflo-component-factory.js
glenna-m, line 85: You are right.  I must have been mixing up the versions of this PR.  Ignore comment above.

test/noflo-component-factory-mocha.js
```diff
@@ -145,7 +116,55 @@ describe('noflo-component-factory', function() {
...
+                    if (err) return reject(err);
+                    network.start();
+                    resolve(network);
+                });
+            }, true);
+        }).then(function(network){
+            return node.nodeName;
+        }).should.eventually.eql(instanceId);
+    });
+    it("should has a componentName", function() {
```
dbooth-boston: s/has/have/

dbooth-boston:
LGTM

catch-point:
I don't think we need a test on isMulti here

glenna-m:
LGTM

PR #49, opened 2016-02-29
Vni manager

glenna-m:
Please review vni-manager.js & its associated test file vni-manager-mocha.js


src/input-states.js
```diff
@@ -0,0 +1,242 @@
...
+ * @param state a State object containing both a data payload and an lm.
+ *
+ * @return getter returns the current input port state state.
+ *         setter returns the current context
+ *
+ * Usage to retrieve state:
+ *  inputState(node, vnid) : {portName: currentState, addressablePortName: [currentState]}
+ *  inputState(node, vnid, portName, socketId) : currentState
+ * Usage to store state:
+ *  inputState(node, vnid, {portName: currentState, addressablePortName: [currentState]}) : this
```
dbooth-boston: Please add a JSdoc comment about @this: will it be the node or the inputStates facade?

src/input-states.js
```diff
@@ -0,0 +1,242 @@
...
+ * @param vnid an identifier that distinguishes the set of VNI states
+ * @param portName name of the input port whose state is being recorded
+ * @param socketId index of the socket or undefined
+ * @param state a State object containing both a data payload and an lm.
+ *
+ * @return getter returns the current input port state state.
+ *         setter returns the current context
+ *
+ * Usage to retrieve state:
+ *  inputState(node, vnid) : {portName: currentState, addressablePortName: [currentState]}
```
dbooth-boston, line 18: Let's use the term 'multi' instead of 'addressable' in RDF Pipeline code (though NoFlo uses the term 'addressable').

Why is this only for a multi port?

src/input-states.js
```diff
@@ -0,0 +1,242 @@
...
+
+/**
+ * Returns a hash of port names to port states or array of states if the port is
+ * addressable/multi. This provides the caller with all portNames, distingishes between
+ * addressable/multi and non-addressable/multi ports, and provides the caller with all their
+ * states that will be useful when calling an updater.
+ * @param node a node facade
+ * @param vnid an identifier that distinguishes the set of VNI states
+ */
+function getAllPortStates(node, vnid) {
```
dbooth-boston, line 46: good

src/input-states.js
```diff
@@ -0,0 +1,242 @@
...
+
+/**
+ * Returns the state of the port or an array of states, if the port is
+ * addressable/multi. When all the port states are not need this is a convient way to
+ * retrieve just the port state that is needed.
+ * @param node a node facade
+ * @param vnid an identifier that distinguishes the set of VNI states
+ * @param portName the InPort name on this node 
+ */
+function getPortState(node, vnid, portName) {
```
dbooth-boston, line 61: good

src/input-states.js
```diff
@@ -0,0 +1,242 @@
...
+
+/**
+ * Returns an array of the port states, one item for each attached socket. When
+ * all the port states are not need this is a convient way to
+ * retrieve just the port state that is needed.
+ * @param node a node facade
+ * @param vnid an identifier that distinguishes the set of VNI states
+ * @param portName the InPort name on this node 
+ */
+function getPortStateArray(node, vnid, portName) {
```
dbooth-boston, line 79: good

src/input-states.js
```diff
@@ -0,0 +1,242 @@
...
+ * look up the last state when checking to see if the state has changed.
+ * Note that the socketId is not neccessarily the same as the array index
+ * @param node a node facade
+ * @param vnid an identifier that distinguishes the set of VNI states
+ * @param portName the name of the addressable/multi port for this node
+ * @param socketId the socketId provided to a port data event
+ */
+function getPortStateBySocketId(node, vnid, portName, socketId) {
+    if (socketId == null)
+        return getPortState(node, vnid, portName);
```
dbooth-boston, line 98: I think it would be safer to throw an error in this case.  The caller should have called getPortState if there is no socketId.

src/input-states.js
```diff
@@ -0,0 +1,242 @@
...
+}
+
+/**
+ * Changes the state of each given port. If a port name is missing it is unmodified.
+ * This can be run in a test environment to reset the state of the ports.
+ * @param node a node facade
+ * @param vnid an identifier that distinguishes the set of VNI states
+ * @param portStates hash of portName to state
+ */
+function setAllPortStates(node, vnid, portStates) {
```
dbooth-boston, line 116: good, i like it

src/input-states.js
catch-point, line 98: I want to make it easy for the caller. If the port is not multi then socketId is undefined. Rather then requiring the caller to check for null socketId and have two different functions, I think it would be easier to do that here and just have one public function for both cases.

src/input-states.js
jamesrdf: okay, but "this" is not used in this function, so it doesn't matter what it is.

src/input-states.js
```diff
@@ -0,0 +1,242 @@
...
+    return _.isFunction(port.isMulti);
+}
+
+/**
+ * Checks if a state already exists for this vnid and portName
+ * @param node a node facade
+ * @param vnid an identifier that distinguishes the set of VNI states
+ * @param portName name of the port on this node
+ */
+function ensureVniStateExists(node, vnid, portName) {
```
dbooth-boston, line 238: Rename ensureVniStateExists can just be renamed to inputStateExists, since it is only checking to see if the state exists.

src/input-states.js
catch-point, line 18: Only multi ports will have an array of states, non-multi ports will have just a state.

src/vni-manager.js
```diff
@@ -0,0 +1,152 @@
...
+ *     var context = component.deleteVnis();
+ *     var context = component.deleteVni( vnid );
+ *     var context = nodeInstance.vni.delete();
+ *     var context = nodeInstance.vni().errorState( state );
+ *     var state = nodeInstance.vni().errorState();
+ *     var context = nodeInstance.vni().inputStates( {input: state} );
+ *     var inputStates = nodeInstance.vni().inputStates();
+ *     var context = nodeInstance.vni().outputState( state );
+ *     var state = nodeInstance.vni().outputState();
+ *     var context = nodeInstance.vni.delete();
```
catch-point: What's this?

src/vni-manager.js
```diff
@@ -0,0 +1,152 @@
...
+    _.extend( facade, 
+        {
+
+          /**
+           * Delete all vnis assiciated with the component node instance.
+           *
+           * @this component context
+           * @return component node instance context for easy chaining
+           */ 
+          deleteVnis: function() {
```
catch-point: What about deleteAllVnis? Sometimes the 's' can be overlooked and node.deleteVnis(vnid) might not do what one expects.

src/input-states.js
```diff
@@ -0,0 +1,242 @@
...
+}
+
+/**
+ * Changes the states of a port.
+ * @param node a node facade
+ * @param vnid an identifier that distinguishes the set of VNI states
+ * @param portName name of the port on this node
+ * @param stateArray a list of states for every attached socket on this port
+ */
+function setPortStateArray(node, vnid, portName, stateArray) {
```
dbooth-boston, line 163: Good.  This would normally only be used in test code though, because events (and states) are received one at a time -- not as an entire array at once.  It might be good to add  comment to that effect.

src/vni-manager.js
```diff
@@ -0,0 +1,152 @@
...
+ * @this the component node instance unless otherwise specified on an API
+ * @param facade a component facade
+ * @return the component facade, extended to include the vni interface
+ * 
+ * Usage examples: 
+ *     var vni = this.nodeInstance.vni(); // retrieve an IIP vni 
+ *     var vni = this.nodeInstance.vni(vnid); // retrieve the vni associated with the vnid
+ *     var context = component.deleteVnis();
+ *     var context = component.deleteVni( vnid );
+ *     var context = nodeInstance.vni.delete();
```
catch-point: This should be var nodeInstance = nodeInstance.vni(vnid).delete(); to distinguish the vni facade context and the nodeInstance facade context.

src/input-states.js
```diff
@@ -0,0 +1,242 @@
...
+function setAllPortStates(node, vnid, portStates) {
+    _.each(portStates, function(state, portName) {
+        setPortState(node, vnid, portName, state);
+    });
+    return node;
+}
+
+/**
+ * Changes the state of a port. When a new state is received on a port this
+ * should be called.
```
dbooth-boston, line 125: s/on a port/on a non-multi port/

And add: setPortStateBySocketId should be used to set the state on a multi/addressable port.

src/input-states.js
```diff
@@ -0,0 +1,242 @@
...
+                if (_.isEmpty(node.vnis[vnid])) {
+                    delete node.vnis[vnid];
+                }
+            }
+        }
+    } else if (port.isMulti()) {
+        return setPortStateArray(node, vnid, portName, state);
+    } else {
+        if (!ensureVniStateExists(node, vnid, portName)) {
+            node.vnis = node.vnis || {};
```
dbooth-boston, line 147: This should call a common ensureVniExists(node, vnid) function instead of creating the object directly, so that all VNI creation is done in one place.

src/input-states.js
```diff
@@ -0,0 +1,242 @@
...
+        throw Error("This port is not addressable/multi: " + portName);
+    port.listAttached().forEach(function(socketId, i) {
+        setPortStateBySocketId(node, vnid, portName, socketId, stateArray[i]);
+    }, []);
+    return node;
+}
+
+/**
+ * Changes the state of a addressable/multi port socket. When a new state is received
+ * on a port this should be called.
```
dbooth-boston, line 175: s/on a port/on a multi port/

And add: setPortState should be called instead if the port is non-multi (non-addressable).

src/input-states.js
```diff
@@ -0,0 +1,242 @@
...
+ * @param node a node facade
+ * @param vnid an identifier that distinguishes the set of VNI states
+ * @param portName name of the port on this node
+ * @param socketId the socketId pass to the data event handler
+ * @param state the new state of this port socket
+ */
+function setPortStateBySocketId(node, vnid, portName, socketId, state) {
+    if (socketId == null)
+        return setPortState(node, vnid, portName, state);
```
dbooth-boston, line 184: throw an error instead.  setPortState should have been called.

src/input-states.js
catch-point, line 125: In a test environment, the port might be a multi port to reset all the states of the port.

src/input-states.js
catch-point, line 184: I think socketId is important in this function, but not important at all to the caller. I think it that would make the caller's code harder to read to move the if statement was move there. It might also require this module to expose another public function (or variation), which we didn't want to do because it might make the module too confusing/busy.

src/input-states.js
dbooth-boston: but the comment above seems to indicate that 'this' is returned

src/input-states.js
dbooth-boston, line 18: oh, i see what you meant.  :)

src/vni-manager.js
```diff
@@ -0,0 +1,152 @@
...
+ * File:  vni-manager.js
+ */
+var _ = require('underscore');
+
+var inputStates = require('./input-states');
+
+var IIP_VNID = '';
+
+/**
+ * Manage the node's vni's by vnid.  If no vnid is provided the IIP VNID will be used.
```
dbooth-boston, line 11: s/IIP VNID/default VNID '' (empty string)/

Explanation: the default VNID is used for regular nodes too -- not only for IIPs.

src/vni-manager.js
```diff
@@ -0,0 +1,152 @@
+/**
+ * File:  vni-manager.js
+ */
+var _ = require('underscore');
+
+var inputStates = require('./input-states');
+
+var IIP_VNID = '';
```
dbooth-boston: s/IIP_VNID/DEFAULT_VNID/g

src/vni-manager.js
```diff
@@ -0,0 +1,152 @@
...
+/**
+ * Manage the node's vni's by vnid.  If no vnid is provided the IIP VNID will be used.
+ * If the vni for the specified vnid does not exist, a new one will be created.
+ *
+ * @this the component node instance unless otherwise specified on an API
+ * @param facade a component facade
+ * @return the component facade, extended to include the vni interface
+ * 
+ * Usage examples: 
+ *     var vni = this.nodeInstance.vni(); // retrieve an IIP vni
```
dbooth-boston: s/retrieve an IIP vni/retrieve the default VNI/

src/vni-manager.js
glenna-m: David had a requirement that we be able to delete a VNI.  I have provided two APIs to do it, though both share the same code underneath.  This one will delete the vni from the vnis array.

src/vni-manager.js
glenna-m: Either is OK with me.  @dbooth-boston  - do you have a preference for the naming of this function:  deleteAllVnis or deleteVnis?

src/vni-manager.js
glenna-m: OK.  Will make that change.

src/vni-manager.js
```diff
@@ -0,0 +1,152 @@
...
+ * If the vni for the specified vnid does not exist, a new one will be created.
+ *
+ * @this the component node instance unless otherwise specified on an API
+ * @param facade a component facade
+ * @return the component facade, extended to include the vni interface
+ * 
+ * Usage examples: 
+ *     var vni = this.nodeInstance.vni(); // retrieve an IIP vni 
+ *     var vni = this.nodeInstance.vni(vnid); // retrieve the vni associated with the vnid
+ *     var context = component.deleteVnis();
```
dbooth-boston: This should be node.deleteAllVnis() -- not component.deleteAllVnis() -- because the VNIs are associated with the node (instances), not the component (factory).

dbooth-boston: s/context/node/ and s/component/node/  -- both are node (instances) AFAICT.  

In general I think it would be cleared to be more specific about what context is returned, such as "node" or "componentFacade" rather than just calling it "context".

src/vni-manager.js
dbooth-boston: And s/nodeInstance/node/

src/vni-manager.js
glenna-m: It is the node instance we are working with.  Will rename.

src/vni-manager.js
```diff
@@ -0,0 +1,152 @@
...
+
+var inputStates = require('./input-states');
+
+var IIP_VNID = '';
+
+/**
+ * Manage the node's vni's by vnid.  If no vnid is provided the IIP VNID will be used.
+ * If the vni for the specified vnid does not exist, a new one will be created.
+ *
+ * @this the component node instance unless otherwise specified on an API
```
dbooth-boston: General comment: There are several places in the code where I found the names confusing.  I think the naming throughout this file should be aligned with the naming conventions that we decided last week:
https://docs.google.com/document/d/1mb_8DCV7nnvhYwc0Ns_gCWpzyXcM1_8qs2u9GJpIZP4/edit#
- An instance of a NoFlo component will be called a "node" (not nodeInstance);
- A NoFlo component (actually a node factory) will be called a "component".  
  - I don't know if we decided what to call the component facade, but how about "componentFacade"?

src/vni-manager.js
```diff
@@ -0,0 +1,152 @@
...
+ */
+module.exports = function( facade ) { 
+
+    _.extend( facade, 
+        {
+
+          /**
+           * Delete all vnis assiciated with the component node instance.
+           *
+           * @this component context
```
dbooth-boston: s/component/node/

src/vni-manager.js
glenna-m: Did not see this before my last commit.  So if I understand write, you want the following mappings:
nodeInstance -> node
facade -> componentFacade.  

If so, will do that now.

dbooth-boston: yes, and avoid using the word "component" when talking about the node instance.  Use "node" instead.  thanks

src/vni-manager.js
```diff
@@ -0,0 +1,155 @@
...
+/**
+ * Manage the node's vni's by vnid.  If no vnid is provided the IIP VNID will be used.
+ * If the vni for the specified vnid does not exist, a new one will be created.
+ *
+ * @this the component node instance unless otherwise specified on an API
+ * @param facade a component facade
+ * @return the component facade, extended to include the vni interface
+ * 
+ * Usage examples: 
+ *     var vni = this.nodeInstance.vni(); // retrieve the default VNI
```
dbooth-boston: AFAICT this should be:

<pre>
var vni = node.vni(); // retrieve the default VNI 
</pre>

src/vni-manager.js
```diff
@@ -0,0 +1,155 @@
...
+ * Manage the node's vni's by vnid.  If no vnid is provided the IIP VNID will be used.
+ * If the vni for the specified vnid does not exist, a new one will be created.
+ *
+ * @this the component node instance unless otherwise specified on an API
+ * @param facade a component facade
+ * @return the component facade, extended to include the vni interface
+ * 
+ * Usage examples: 
+ *     var vni = this.nodeInstance.vni(); // retrieve the default VNI 
+ *     var vni = this.nodeInstance.vni(vnid); // retrieve the vni associated with the vnid
```
dbooth-boston: Similarly this should be:

<pre>
var vni = node.vni(vnid); // retrieve the vni associated with the vnid
</pre>

src/vni-manager.js
```diff
@@ -0,0 +1,155 @@
...
+ *
+ *     component.deleteAllVnis();
```
dbooth-boston: s/component/node/ and in several more places.

src/vni-manager.js
```diff
@@ -0,0 +1,155 @@
...
+ *     component.deleteVni( vnid );
+ *     nodeInstance.vni(vnid).delete();
```
dbooth-boston: s/nodeInstance/node/  and similarly in several more places.

src/vni-manager.js
```diff
@@ -0,0 +1,155 @@
...
+    _.extend( facade, 
+        {
+
+          /**
+           * Delete all vnis assiciated with the node instance.
+           *
+           * @this node instance
+           * @return node instance nodeInstance for easy chaining
+           */ 
+          deleteAllVnis: function() {
```
dbooth-boston, line 47: good

src/vni-manager.js
```diff
@@ -0,0 +1,156 @@
...
+
+              return {
+                  delete: _.bind( this.deleteVni, this, vnid ),
+                  inputStates: _.partial( inputStates, this, vnid ), 
+                  errorState: _.partial( errorState, this.vnis[vnid] ), 
+                  outputState: _.partial( outputState, this.vnis[vnid] ) 
+              };
+          },
+
+          vnis: {}
```
dbooth-boston, line 100: I am confused about whether this module returns a _component_ facade or a _node_ (instance).  The vnis hash needs to be stored on the _node_ -- not the component facade.  Maybe you can explain on a hangout.

src/vni-manager.js
```diff
@@ -0,0 +1,156 @@
...
+    }); // extend
+} // module export
+
+/**
+ * Get/Set error state on the vni 
+ * 
+ * @this vni context 
+ * @param state error state to be set
+ */
+function errorState( vni, state ) {
```
dbooth-boston, line 115: good

src/vni-manager.js
```diff
@@ -0,0 +1,156 @@
...
+}
+
+/**
+ * Get/Set output state on the vni 
+ * 
+ * @this vni context 
+ * @param state output state to be set
+ */
+function outputState( vni, state ) {
+
```
dbooth-boston, line 146: good

test/vni-manager-mocha.js
```diff
@@ -0,0 +1,306 @@
...
+
+var test = require('./common-test');
+
+var vniManager = require('../src/vni-manager');
+
+describe('vni-manager', function() {
+
+    beforeEach(function() {
+
+        component = test.createComponent(componentFactory({
```
dbooth-boston: s/component/node/

test/vni-manager-mocha.js
```diff
@@ -0,0 +1,306 @@
...
+        component.deleteAllVnis();
+    });
+
+    it('should exist as a function', function() {
+      vniManager.should.exist;
+      vniManager.should.be.a('function');
+    });
+
+    describe('#vnis', function() {
+        it('should exist as an object in the component instance', function() {
```
dbooth-boston: s/component instance/node (instance)/ and throughout this file.

test/vni-manager-mocha.js
```diff
@@ -0,0 +1,306 @@
...
+            component.vnis.should.exist;
+            component.vnis.should.be.a('object');
+            Object.keys( component.vnis ).should.have.length( 0 );
+        });
+    });
+
+    describe('#vni', function() {
+
+        it('should create and return a new IIP vni if no vnid is specified and no vni exists', function() {
```
dbooth-boston: s/IIP/default/

test/vni-manager-mocha.js
```diff
@@ -0,0 +1,306 @@
...
+            testVni.should.be.an('object');
+            testVni.inputStates.should.exist;
+            testVni.inputStates.should.be.a('function');
+            testVni.errorState.should.exist;
+            testVni.errorState.should.be.a('function');
+            testVni.outputState.should.exist;
+            testVni.outputState.should.be.a('function');
+
+            expect( testVni.parentVni ).to.be.undefined;
+            expect( testVni.previousLms ).to.be.undefined;
```
dbooth-boston: The previousLms would be a property of the output or error states -- not directly in the VNI.

test/vni-manager-mocha.js
```diff
@@ -0,0 +1,306 @@
...
```
dbooth-boston: ditto

test/vni-manager-mocha.js
```diff
@@ -0,0 +1,306 @@
...
+            it('should delete the VNI', function() {
+
+                Object.keys( component.vnis ).should.have.length( 0 );
+
+                var testVnid = '';
+                var testVni = component.vni( testVnid );
+                testVni.should.be.an('object');
+                Object.keys( component.vnis ).should.have.length( 1 );
+
+                var context = testVni.delete();
```
dbooth-boston: s/context/node/   and several other places below

test/vni-manager-mocha.js
```diff
@@ -0,0 +1,306 @@
...
+                var context = testVni.outputState( state );
+                context.should.be.an('object');
+
+                // Test clearing the state 
+                var clearContext = testVni.outputState( undefined );
+                clearContext.should.be.an('object');
+
+                // Test get of cleared state 
+                var clearedState = testVni.outputState();
+                expect( clearedState ).to.be.an('undefined');
```
dbooth-boston, line 319: good

src/vni-manager.js
glenna-m, line 100: Added a test that demonstrates that the function results in an extended component facade with the new vni-manager apis.

test/vni-manager-mocha.js
```diff
@@ -0,0 +1,377 @@
...
+                    ondata: promiseOutput(function(payload){
+                        return payload + " world2";
+                    })
+                }},
+                outPorts: promiseOutput.outPorts
+              }, vniManager )
+            );
+ 
+            node1.should.not.equal(node2);
+            node1.vnis.should.not.equal(node2.vnis);
```
dbooth-boston, line 84: good, thanks

dbooth-boston:
LGTM.  thanks!

src/vni-manager.js
```diff
@@ -0,0 +1,158 @@
...
+        if ( _.isUndefined(state) ) { 
+            // Clear state
+            delete vni.errorState;
+        } else { 
+            // Set state
+            vni.errorState = state;
+        }
+
+        return node;
```
catch-point: I think this should be 'return this;' so that it returns the vni facade and not the node facade. Same goes for outputState below.

src/vni-manager.js
```diff
@@ -0,0 +1,158 @@
...
+
+              vnid = vnid || DEFAULT_VNID;
+
+              if ( _.isUndefined( this.vnis[vnid] )  ) {
+
+                  // have no vni so create one
+                  this.vnis[vnid] = {}; 
+
+                  // TODO: Add parentVni setting here
+                  // TODO: Add previousLms here
```
dbooth-boston: Not previousLms.  They are a part of the state objects -- not the VNI directly.

test/vni-manager-mocha.js
```diff
@@ -0,0 +1,377 @@
...
+
+var test = require('./common-test');
+
+var vniManager = require('../src/vni-manager');
+
+describe("vni-manager", function() {
+
+    beforeEach(function() {
+
+        node = test.createComponent(componentFactory({
```
dbooth-boston, line 24: Should we rename createComponent to createNode?  I'm getting confused about which objects are components (i.e., node factories) and which are nodes (i.e., node instances).

catch-point:
LGTM

test/vni-manager-mocha.js
```diff
@@ -0,0 +1,377 @@
...
+        // clean up
+        node.deleteAllVnis();
+    });
+
+    it("should exist as a function", function() {
+        vniManager.should.exist;
+        vniManager.should.be.a('function');
+    });
+
+    it("should return a component facade", function() {
```
dbooth-boston, line 46: Should this say "node facade" instead of "component facade"?  I'm still a bit confused about which objects are nodes and which are components, so if we can be consistent in our naming it will help.

test/vni-manager-mocha.js
```diff
@@ -0,0 +1,377 @@
+/**
+ * File: vni-manager-mocha.js
+ * Unit tests for the vni APIs defined in src/vni-manager.js
+ */
+
+var chai = require('chai');
+var expect = chai.expect;
+var should = chai.should();
+
+var componentFactory = require('../src/noflo-component-factory');
```
dbooth-boston, line 10: Let's rename noflo-component-factory --> noflo-node-factory
and componentFactory --> nodeFactory

But we can do it in a separate PR

dbooth-boston:
LGTM!

test/vni-manager-mocha.js
catch-point, line 46: My take away from yesterdays discussion on this is that the term node should be used when referencing a particular instance and the term component should be used when referencing a class/collection of them.

PR #50, opened 2016-03-02
Framework

glenna-m:
Three files for review in this PR (the others are from other PRs): 
- framework.js
- framework-mocha.js
- common-test.js

### DIFF
--- a/src/framework.js
+++ b/src/framework.js
@@ -1,0 +1,143 @@
+/**
+ * File: framework.js 
+ */
+var _ = require('underscore');
+var util = require('util');
+
+var componentFactory = require('./noflo-component-factory');
+var stateFactory = require('./create-state');
+var promiseOutput = require('../src/promise-output');
+var vniManager = require('./vni-manager');
+
+/**
+ * This module implements the core of the RDF Pipeline Framework.  
+ */
+module.exports = {
+
+    componentFactory: function( nodeDef, fRunUpdater ) { 
+        
+      return componentFactory(
+          _.defaults( { // Add an ondata function to be invoked every time we receive new data
+                        inPorts: _.mapObject(nodeDef.inPorts, function( port, portName ) {  
+                            return _.extend({
+                                ondata: _.partial( ondata, fRunUpdater ) 
+                            }, port);
+                        }),
+                        outPorts: _.extend( {},
+                                            promiseOutput.outPorts,
+                                            nodeDef.outPorts )
+                      },
+                      nodeDef
+          ), 
+          vniManager 
+      );
+    }
+
+}; // module.exports
+
+
+// @this port context
+var ondata = function( fRunUpdater, payload, socketIndex ) {
+
+     var portName = this.name;
+
+     var vnid = payload.vnid || '';
+     var vni = this.nodeInstance.vni(vnid);
+
+     // Check if we have a multi/addressable port and call the correct input-state API
+     if ( _.isUndefined( socketIndex ) ) { 
+         vni.inputStates( { [portName] : stateFactory( vnid, payload ) });
+     } else {
+         vni.inputStates( portName, socketIndex, stateFactory( vnid, payload ) );
+     }
+
+     // check if should run updater & run if necessary
+     if ( shouldRunUpdater( this.nodeInstance, vni ) ) { 
+          
+        // Save vars we will need in the promise where the context is different
+        var outputPorts = this.nodeInstance.outPorts;
+        var errorPorts = this.nodeInstance.errorPorts;
+
+        var promise = fRunUpdater.call( this.nodeInstance, vni.inputStates() );
+        Promise.resolve( promise ).then( function( data ) { 
+
+            // Set output state and send it on over the output port
+            if (data) {
+                var outputState = stateFactory( vnid, data );
+                vni.outputState( outputState );
+
+                // Should this be in access or output layer? 
+                outputPorts.output.send( outputState );
+                outputPorts.output.disconnect();
+            }
+        }, function( error ) { 
+
+            // Set error state and send it on over the error port
+            var errorState = stateFactory( vnid, error );
+            vni.errorState( errorState );
+
+            errorPorts.output.send( errorState );
+            errorPorts.output.disconnect();
+        }); 
+     } 
+
+     return;
+};
+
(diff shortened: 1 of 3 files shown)


src/framework.js
```diff
@@ -0,0 +1,143 @@
...
+        for ( var i=0, max=requiredInPorts.length; i < max; i++ ) { 
+
+            var port = requiredInPorts[i];
+            var states = vni.inputStates( port.name );
+            
+            // figure out how many input states we have received
+            var numberOfStates = 0;
+            if ( _.isArray( states ) ) { 
+               // filter out any undefined states and then count what we really have
+               var flattenedStates = _.filter( states, function(state) {
```
catch-point, line 112: _.compact does just this

src/framework.js
```diff
@@ -0,0 +1,143 @@
...
+     } else {
+         vni.inputStates( portName, socketIndex, stateFactory( vnid, payload ) );
+     }
+
+     // check if should run updater & run if necessary
+     if ( shouldRunUpdater( this.nodeInstance, vni ) ) { 
+          
+        // Save vars we will need in the promise where the context is different
+        var outputPorts = this.nodeInstance.outPorts;
+        var errorPorts = this.nodeInstance.errorPorts;
```
catch-point, line 59: This line is wrong and needs to be tested

src/framework.js
```diff
@@ -0,0 +1,143 @@
...
+// @this port context
+var ondata = function( fRunUpdater, payload, socketIndex ) {
+
+     var portName = this.name;
+
+     var vnid = payload.vnid || '';
+     var vni = this.nodeInstance.vni(vnid);
+
+     // Check if we have a multi/addressable port and call the correct input-state API
+     if ( _.isUndefined( socketIndex ) ) {
```
catch-point, line 48: No need to do this the input-states.js will handle an undefined socketId, just use full form.

src/framework.js
```diff
@@ -0,0 +1,143 @@
...
+var stateFactory = require('./create-state');
+var promiseOutput = require('../src/promise-output');
+var vniManager = require('./vni-manager');
+
+/**
+ * This module implements the core of the RDF Pipeline Framework.  
+ */
+module.exports = {
+
+    componentFactory: function( nodeDef, fRunUpdater ) {
```
dbooth-boston, line 17: The framework will need access to the whole wrapper -- not only its fRunUpdater.  Also, I'm now wondering if (some of?) this should be in the noflo access layer.  We should discuss this on a design teleconference, since it is uncharted territory.

test/framework-mocha.js
```diff
@@ -0,0 +1,420 @@
...
+                                  'vnis', 'deleteAllVnis', 'deleteVni', 'vni' );
+
+        // Verify we got the ports we expect to have
+        Object.keys( node.inPorts ).should.have.length(1);
+        node.inPorts.should.have.all.keys( 'input1' );
+        Object.keys( node.outPorts ).should.have.length(2);
+        node.outPorts.should.have.all.keys( 'output', 'error' );
+    });
+
+    it( "should support creation of node instances with custom output ports", function() {
```
dbooth-boston, line 73: Do we need custom output ports?  If not I think we should drop this test.

test/framework-mocha.js
```diff
@@ -0,0 +1,420 @@
...
+                        datatype: 'string',
+                        description: "a string port",
+                        required: true
+                    }
+                }
+            }, fRunUpdater));
+
+            // Send data to the input port and verify that the fRunUpdater function is called.
+            // We use a promise here because this section is asynchronous
+            return new Promise( function(done, fail) {
```
dbooth-boston, line 136: Suggest renaming these parameters to be resolve and reject, in accordance with normal Promises convention, and to distinguish from the 'done' and 'fail' parameters below, which are (I believe) completely different.

test/framework-mocha.js
```diff
@@ -0,0 +1,420 @@
...
+
+            // Send data to the input port and verify that the fRunUpdater function is called.
+            // We use a promise here because this section is asynchronous
+            return new Promise( function(done, fail) { 
+                test.onOutPortData(node, 'output', done);
+                test.onOutPortData(node, 'error', fail);
+                test.sendData(node, 'input1', inData);
+            }).then( function( done ) { 
+               done.vnid.should.equal( '' ); 
+               done.data.should.equal( executedFRunUpdater );
```
dbooth-boston, line 142: good

test/framework-mocha.js
```diff
@@ -0,0 +1,420 @@
...
+            var node = test.createComponent( framework.componentFactory({
+                description: "Test Description",
+                inPorts: {
+                    input1: {
+                        datatype: 'string',
+                        description: "a string port",
+                        required: true
+                    }
+                }
+            }, fRunUpdater));
```
dbooth-boston, line 132: We should discuss how the fRunUpdater should be provided.

test/framework-mocha.js
```diff
@@ -0,0 +1,420 @@
...
+        });
+
+        it( "should manage multiple port node vni state, fRunUpdater invocation, & output state", function() {
+
+            var input1 = 'Uno';
+            var input2 = 'Dos';
+            var executedFRunUpdater = "Executed fRunUpdater API";
+
+            // Define the fRunUpdater that framework should invoke
+            var fRunUpdater = function( inputStates ) {
```
dbooth-boston, line 155: fRunUpdater needs to have access to the entire VNI -- not only the inputStates.

src/framework.js
glenna-m, line 48: OK - will be in the next checkin

src/framework.js
```diff
@@ -0,0 +1,143 @@
...
+     }
+
+     // check if should run updater & run if necessary
+     if ( shouldRunUpdater( this.nodeInstance, vni ) ) { 
+          
+        // Save vars we will need in the promise where the context is different
+        var outputPorts = this.nodeInstance.outPorts;
+        var errorPorts = this.nodeInstance.errorPorts;
+
+        var promise = fRunUpdater.call( this.nodeInstance, vni.inputStates() );
```
catch-point, line 61: fRunUpdater should be called within a Promise to catch any exceptions that it produces directly.

PR #53, opened 2016-03-04
Add rdf-insert Component

catch-point:
This takes jsonld and inserts it into the given SPARQL endpoint within the given graph

### DIFF
--- a/components/rdf-ntriples.js
+++ b/components/rdf-ntriples.js
@@ -27,6 +27,7 @@ exports.getComponent = componentFactory({
  * Promise of N-Triples array of tokens for the given graph.
  * @this a noflo.InPort or facade
  * @param graph RDF JS Interface Graph object
+ * @see https://www.w3.org/TR/rdf-interfaces/#graphs
  */
 function execute(graph) {
     var tokens = [];
@@ -43,7 +44,8 @@ function execute(graph) {
 
 /**
  * Converts the RDF term into a array of tokens
- * @param object an RDF term implementing RDF JS Interface
+ * @param term an RDF term implementing RDF JS Interface
+ * @see https://www.w3.org/TR/rdf-interfaces/#basic-node-types
  */
 function valueTokens(term) {
     var value = term.nominalValue;
@@ -73,9 +75,9 @@ function valueTokens(term) {
 /**
  * Checks if the string includes any special characters.
  */
-function encodingNeededFor(st) {
-    return st.indexOf('\\') >= 0 || st.indexOf('\"') >= 0 ||
-        st.indexOf('\n') >= 0 || st.indexOf('\r') >= 0 || st.indexOf('\t') >= 0;
+function encodingNeededFor(string) {
+    return string.indexOf('\\') >= 0 || string.indexOf('\"') >= 0 ||
+        string.indexOf('\n') >= 0 || string.indexOf('\r') >= 0 || string.indexOf('\t') >= 0;
 }
 
 /**
--- a/components/rdf-ntriples.js
+++ b/components/rdf-ntriples.js
@@ -1,0 +1,86 @@
+// rdf-ntriples.js
+
+var _ = require('underscore');
+
+var promiseOutput = require('../src/promise-output');
+var componentFactory = require('../src/noflo-component-factory');
+
+/**
+ * Converts an RDF JS Interface Graph object into an object with the property
+ * 'tokens' containing an Array of N-Triples tokens. tokens.join('') will produce an N-Triples string.
+ */
+exports.getComponent = componentFactory({
+    description: "Converts an RDF JS Interface Graph object into an Array of N-Triples tokens",
+    icon: 'edit',
+    outPorts: promiseOutput.outPorts,
+    inPorts: {
+        input: {
+            description: "RDF JS Interface Graph object",
+            datatype: 'object',
+            required: true,
+            ondata: promiseOutput(execute)
+        }
+    }
+});
+
+/**
+ * Promise of N-Triples array of tokens for the given graph.
+ * @this a noflo.InPort or facade
+ * @param graph RDF JS Interface Graph object
+ */
+function execute(graph) {
+    var tokens = [];
+    graph.forEach(function(triple) {
+        tokens.push.apply(tokens, valueTokens(triple.subject));
+        tokens.push(' ');
+        tokens.push.apply(tokens, valueTokens(triple.predicate));
+        tokens.push(' ');
+        tokens.push.apply(tokens, valueTokens(triple.object));
+        tokens.push(' .', '\n');
+    });
+    return {tokens: tokens};
+}
+
+/**
+ * Converts the RDF term into a array of tokens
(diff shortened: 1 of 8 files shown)


dbooth-boston:
LGTM

components/rdf-ntriples.js
```diff
@@ -0,0 +1,86 @@
...
+var componentFactory = require('../src/noflo-component-factory');
+
+/**
+ * Converts an RDF JS Interface Graph object into an object with the property
+ * 'tokens' containing an Array of N-Triples tokens. tokens.join('') will produce an N-Triples string.
+ */
+exports.getComponent = componentFactory({
+    description: "Converts an RDF JS Interface Graph object into an Array of N-Triples tokens",
+    icon: 'edit',
+    outPorts: promiseOutput.outPorts,
```
glenna-m, line 15: We need to remember to remove this when you use the wrapper/framework - it is done for you there..

components/rdf-ntriples.js
```diff
@@ -0,0 +1,86 @@
...
+        tokens.push(' ');
+        tokens.push.apply(tokens, valueTokens(triple.object));
+        tokens.push(' .', '\n');
+    });
+    return {tokens: tokens};
+}
+
+/**
+ * Converts the RDF term into a array of tokens
+ * @param object an RDF term implementing RDF JS Interface
```
glenna-m: parameter name does not match the actual parameter (term). That's a nit, but more interesting would be a bit more documentation on the term object structure - it looks to me like it's coming from some kind of possibly standard source.  If so, can you point to that?

components/rdf-ntriples.js
```diff
@@ -0,0 +1,86 @@
...
+    } else {
+        return ['"', encoded, '"'];
+    }
+}
+
+/**
+ * Checks if the string includes any special characters.
+ */
+function encodingNeededFor(st) {
+    return st.indexOf('\\') >= 0 || st.indexOf('\"') >= 0 ||
```
glenna-m: you spell it out as string rather than st most other places - can we keep with that so it's a no-brainer at first glance as to what that parameter is?

graphs/rdf-insert.fbp
```diff
@@ -0,0 +1,27 @@
+# rdf-insert.fbp
+
+# Takes parsed JSON-LD and inserts it into the target-graph (IRI) at the sparql-endpoint.
```
glenna-m, line 3: Based on the discussion this morning, I thought there was going to be a bit more documentation here?

test/common-test.js
```diff
@@ -24,7 +24,7 @@ module.exports = {
         var graph = new noflo.Graph();
         _.each(componentModules, function(module, name) {
```
glenna-m: Can we use a more descriptive name than "name" ?  This is the component name, yes?   Can we call this something like "componentName"?

test/rdf-components-mocha.js
```diff
@@ -49,6 +52,21 @@ describe('rdf components', function() {
             });
         }).should.eventually.include.keys('triples');
     });
+    it("should round trip a graph", function() {
+        return test.createNetwork({
```
glenna-m, line 57: Can we add some comments being really clear what's being done?  If I understand right, I think we are creating a 2 node graph here - one with the rdfLoad component and one with rdfJsonld - correct?    A comment like 
    // creating 2 node graph (rdfLoad & rdfJsonld components)

test/rdf-components-mocha.js
```diff
@@ -49,6 +52,21 @@ describe('rdf components', function() {
...
+            load: rdfLoad,
+            jsonld: rdfJsonld
+        }).then(function(network){
```
glenna-m, line 60: Then same thing here - a single line comment that explains the block

test/rdf-components-mocha.js
```diff
@@ -104,4 +122,30 @@ describe('rdf components', function() {
             });
         }).should.eventually.eql(cynthia);
     });
+    it("should round trip a graph through ntriples", function() {
```
glenna-m: Not sure I understand this test description.  Round tripping a graph by sending data, I get.  But what is round tripping through ntriples?  Aren't the ntriples the data?

test/rdf-insert-mocha.js
```diff
@@ -0,0 +1,85 @@
...
+            return new Promise(function(done) {
+                output.on('data', done);
+                network.graph.addInitial("http://localhost:1337/", 'insert', 'sparql_endpoint');
+                network.graph.addInitial(john, 'insert', 'jsonld');
+            }).then(function(sparql){
+                return _.isString(sparql) ? sparql.replace(/\s+/g,'\n').trim() : sparql;
+            });
+        }).should.become(sparql.replace(/\s+/g,'\n').trim());
+    });
+});
```
glenna-m, line 86: This is all inserts.  Do we need to support/test updates too?

test/rdf-insert-mocha.js
```diff
@@ -0,0 +1,85 @@
...
+        var body = [];
+        req.on('data', function(chunk) {
+            body.push(chunk);
+        }).on('end', function() {
+            body = Buffer.concat(body).toString();
+            res.end(body);
+        });
+    });
+    before(function(){
+        server.listen(1337);
```
glenna-m: Interesting choice of a port #.  Is there a reason you chose 1337?   If so, please document it.

test/rdf-insert-mocha.js
```diff
@@ -0,0 +1,85 @@
...
+    });
+    it("should POST jsonld as SPARQL INSERT using the default graph", function() {
+        return test.createNetwork({
+            insert: "rdf-components/rdf-insert"
+        }).then(function(network){
+            var output = noflo.internalSocket.createSocket();
+            network.processes.insert.component.outPorts.output.attach(output);
+            return new Promise(function(done) {
+                output.on('data', done);
+                network.graph.addInitial("http://localhost:1337/", 'insert', 'sparql_endpoint');
```
glenna-m: Since the port number is being used multiple places, can we put it into a variable so it can be changed if we want to?

graphs/rdf-insert.fbp
catch-point, line 3: Like what?

test/rdf-components-mocha.js
catch-point: Round trips the data, like the previous test, but this goes through an rdf-ntriples component an its way.

test/rdf-insert-mocha.js
catch-point: It could have been anything, I choose this number bc it wasn't 8080, which is more commonly used for services.

test/rdf-insert-mocha.js
catch-point, line 86: Maybe, but not in the rdf-insert component ;)

test/common-test.js
catch-point: In this context the same name is used for both nodeId and componentName.

glenna-m:
LGTM

PR #54, opened 2016-03-07
Add the node instance onto the VNI facade

glenna-m:
In review discussions last week, David indicated that we should be calling the updater on a vni context and make the node instance available to the updaters for some of the more sophisticated use cases.  This check in adds the node facade to the vni facade to support this use case.

While testing, I noticed that the input states setter behaviour has changed after some of the recent merging.   The changes are making a set state like this: 
       var result = node.vni().inputStates( {input: state} );
return a VNI facade, where previously they were returning a node facade.   Reviewers should weigh in on what context we really want on a setter return - is this correct?

### DIFF
--- a/src/vni-manager.js
+++ b/src/vni-manager.js
@@ -89,11 +89,13 @@ module.exports = function( componentFacade ) {
                   // TODO: Add parentVni setting here
               }
 
+              var that = this; // save current node context to reference in facade
               return {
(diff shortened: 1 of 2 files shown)


catch-point:
LGTM

dbooth-boston:
LGTM!

PR #58, opened 2016-03-08
framework-ondata.js and unit testing

glenna-m:
Latest version of the framework-ondata function and its supporting code and tests.  

This version does not deal with PreviousLms since we are still trying to complete Sprint 5.  It does deal with a number of other changes David has requested in the past week including executing on the vni context.

### DIFF
--- a/src/create-state.js
+++ b/src/create-state.js
@@ -17,7 +17,7 @@ var createLm = require('./create-lm');
  *
  * @return the newly created state object.
  */ 
-module.exports = function( vnid, data, lm ) { 
+module.exports = function( vnid, data, lm, error ) { 
 
        if ( _.isUndefined( vnid) ) {
            throw new Error("Unable to create state because no vnid was provided");
@@ -26,12 +26,17 @@ module.exports = function( vnid, data, lm ) {
        var stateLm;
        if ( _.isUndefined( data ) && _.isUndefined( lm ) ) { 
            // If both data & lm are undefined, we are initializing - accept an undefined lm
-           stateLm = lm;
-       } else {
-           // Either data or LM is not undefined - so use the lm or create a new one if we do not have an lm
-           stateLm = lm || createLm();
-       }
+           return { vnid: vnid,
+                    data: undefined, 
+                    error: undefined,
+                    lm: undefined };
+
+       } 
+
+       // Either data or LM is not undefined - so use the lm or create a new one if we do 
+       // not have an lm
        return { vnid: vnid,
                 data: data, 
-                lm: stateLm };
+                error: error,
+                lm: lm || createLm() };
 };
--- a/src/create-state.js
+++ b/src/create-state.js
@@ -23,7 +23,14 @@ module.exports = function( vnid, data, lm ) {
            throw new Error("Unable to create state because no vnid was provided");
        }
 
-       var stateLm = lm || createLm();
+       var stateLm;
+       if ( _.isUndefined( data ) && _.isUndefined( lm ) ) { 
+           // If both data & lm are undefined, we are initializing - accept an undefined lm
+           stateLm = lm;
+       } else {
+           // Either data or LM is not undefined - so use the lm or create a new one if we do not have an lm
+           stateLm = lm || createLm();
+       }
        return { vnid: vnid,
                 data: data, 
                 lm: stateLm };
--- a/src/framework-ondata.js
+++ b/src/framework-ondata.js
@@ -28,80 +28,139 @@ module.exports = function( payload, socketIndex ) {
      var vnid = payload.vnid || '';
      var vni = this.nodeInstance.vni(vnid);
 
-     // Update the input state
      var inputState = (payload.vnid) ? payload : stateFactory( vnid, payload );
      vni.inputStates( portName, socketIndex, inputState );
 
-     // check if should run updater & run if necessary
      if ( shouldRunUpdater( vni ) ) { 
-          
-        // Save vars we will need in the promise where the context is different
-        var outputPorts = this.nodeInstance.outPorts;
 
-        if ( _.isUndefined( this.nodeInstance.wrapper.fRunUpdater ) || 
+         // Save vars we will need in the promise where the context is different
+         var outputPorts = this.nodeInstance.outPorts;
+
+         if ( _.isUndefined( this.nodeInstance.wrapper.fRunUpdater ) || 
              ! _.isFunction( this.nodeInstance.wrapper.fRunUpdater ) ) { 
 
              // Don't have a wrapper fRunUpdater 
              throw Error( 'No wrapper fRunUpdater function found!  Cannot run updater.' );
 
-        } else { 
+         } else { 
 
-            // save current error & output state LMs to compare after updater runs
-            var lastOutputLm = vni.outputState().lm;
-            var lastErrorLm = vni.errorState().lm;
-            var wrapper = this.nodeInstance.wrapper;
+             // save data used to see if we have a different state after running updater
+             var lastErrorState = _.clone( vni.errorState() );  // Shallow clone 
+             var lastOutputLm = vni.outputState().lm;
+ 
+             clearStateData( _.partial( vni.errorState ) ); // clear last error state
 
-            new Promise(function( resolve ) { 
-                resolve( wrapper.fRunUpdater.call( vni ) );
+             // Save wrapper to make it accessible from the Promise
+             var wrapper = this.nodeInstance.wrapper;
 
-            }).then( function() { 
-                // If not new output or error state data, send it downstream or log it to the console
-                handlePortOutput( outputPorts.output, lastOutputLm, vni.outputState() );
-                handlePortOutput( outputPorts.error,  lastErrorLm, vni.errorState() );
+             new Promise(function( resolve ) { 
+ 
+                 // Execute fRunUpdater which will also execute the updater
+                 resolve( wrapper.fRunUpdater.call( vni ) );
 
-            }, function( rejected ) { 
-    
-                console.error( rejected );
+             }).then( function() { 
+                 // fRunUpdater/Updater success path
+             
 
-                // fRunUpdater failed - is this a new error? 
-                var errorState = vni.errorState();
+                 // Returned OK, but updater could have set an error - check for that
+                 if ( stateChange( _.partial( vni.errorState ), lastErrorState ) ) {
+                    lastOutputLm = undefined;  // got a new error error so we will force sending output
+                 }
 
-                if ( lastErrorLm !== errorState.lm  ) { 
+                 setOutputErrorFlag( vni );   // set error flag to true if new error, false if not
 
-                    // fRunUpdater or updater recorded a new lm (i.e., new error) - send it on
-                    handlePortOutput( outputPorts.error, lastErrorLm, errorState );
+                 // If not new output or error state data, send it downstream or log it to the console
+                 handleOutput( outputPorts.output, lastOutputLm, vni.outputState() );
+                 handleError( outputPorts.error,  lastErrorState.lm, vni.errorState() );
 
-                 } else if ( rejected !== errorState.data ) { 
+             }, function( rejected ) { 
+                 // fRunUpdater/updater failed 
+                 if ( isInitState( vni.errorState ) ) { 
+                     // fRunUpdater/updater has not already set an error state - use the rejected info
+                     changeStateData( _.partial( vni.errorState ), rejected );
+                 }
 
-                    // Same LM as before we called fRunUpdater, but new error data was detected
-                    // TODO: vni.error.setPreviousLmsFromInputStates(vni);
-                    errorState.lm = createLm();
-                    errorState.data = rejected;
-                    vni.errorState( errorState );
-                    handlePortOutput( outputPorts.error, lastErrorLm, vni.errorState() );
-                 } 
+                 setOutputErrorFlag( vni ); 
 
-            }).catch( function() { 
-               console.error( "framework-ondata unable to process fRunUpdater results!" );
-               reject();
-            } ); 
-            
-        }
-     } 
+                 stateChange( _.partial( vni.outputState ), lastOutputLm );
+                 stateChange( _.partial( vni.errorState ), lastErrorState );
+
+                 handleOutput( outputPorts.output, lastOutputLm, vni.outputState() );
+                 handleError( outputPorts.error, lastErrorState.lm, vni.errorState() );
+
+                 // If we haven't already processed the rejected error, do it now
+                 if ( rejected !== vni.errorState().data ) { 
+                     lastErrorState = _.clone( vni.errorState() );   
+                     changeStateData( _.partial( vni.errorState ), rejected );
+                     if ( changeStateData( _.partial( vni.errorState ), lastErrorState ) ) { 
+                         lastErrorState.lm = undefined; 
+                     } 
+                     handleError( outputPorts.error, lastErrorState.lm, vni.errorState() );
+                 }
+
+             }).catch( function( e ) { 
+                 // fRunUpdater or Updater threw an error
+                 console.error( "framework-ondata unable to process fRunUpdater results!" );
+                 var err = ( e.stack ) ? e.stack : e;
+             }); 
+         } // have an fRunUpdater
+     } // not ready for fRunUpdater yet
 };
 
-/** 
- * Retrieve all attached input ports for the noflo component associated with this
- * node instance. 
- *
- * @this node instance context
- *
- * @return an array with the attached noflo port objects
- */
-function attachedInputPorts() { 
+/*******************************************************************************/
+/* Error & Output State handling                                               */
+/*******************************************************************************/                 
 
-    return _.filter( this.inPorts, 
-                     function( port) { return  port.listAttached().length > 0; });
+/**
+ * change state data to the specified newData value
+ *
+ * @param stateFacade a getter/setterfunction for the current state
+ */ 
+function changeStateData( stateFacade, newData, newLm ) { 
+    var currentState = stateFacade();
+    currentState.data = newData;
+    if ( newLm ) { 
+        currentState.lm = createLm();
+    }
+    stateFacade( currentState );
+}
+
+/**
+ * Clears the state data field by setting it to undefined.  
+ *
+ * @param stateFacade a getter/setterfunction for the current state
+ */
+function clearStateData( stateFacade ) { 
+    changeStateData( stateFacade ); 
+}
+
+/**
+ * Sends an error state on to any attached port on a down stream node if it's a new error.
+ * If there is no attached port, logs the error so it can be analyzed.
+ *
+ * @param port the error port, which may or may not be attached to something down stream
+ * @param lastLm last recorded state lm
+ * @param state an error or output state to be sent down stream or logged
+ */
+function handleError( port, lastLm, state ) { 
+
+    if ( (! handleOutput( port, lastLm, state )) && state.data )  { 
+        // State was not sent on to an attached port, and we do have error data
+        // so go ahead and log it for debugging/support use.
+        if ( state.data.stack ) { 
+            console.error( state.data.stack );
+        } else { 
+            console.error( state.data );
+        }
+    }
+}
+
+/**
+ * Returns true if the state is an initial vni state that has not got any
+ * values in it yet.
+ */
+function isInitState( state ) { 
+    return ( _.isUndefined( state.data ) && _.isUndefined( state.lm ) && _.isUndefined( state.error ) );
 }
 
 /**
@@ -111,8 +170,11 @@ function attachedInputPorts() {
  * @param port an output port, which may or may not be attached to something down stream
  * @param lastLm last recorded state lm
  * @param state an error or output state to be sent down stream or logged
+ *
+ * @return true if the data was sent to an attached port on a down stream node;
+ *         false if the data was not sent on
  */
-function handlePortOutput( port, lastLm, state ) {
+function handleOutput( port, lastLm, state ) {
 
     // Do we have a new state? 
     if ( lastLm !== state.lm ) { 
@@ -121,8 +183,99 @@ function handlePortOutput( port, lastLm, state ) {
         if (port.listAttached().length) {
             port.send( state );
             port.disconnect();
-        } 
+            return true;
+        }  
     } 
+
+    return false;
+}
+
+/**
+ * Set the output state error flag to true if there is a current error, false if not.
+ * 
+ * @param vni 
+ * @param errorFlag (optional) if not undefined, the errorFlag value will be set
+ */
+function setOutputErrorFlag( vni, errorFlag ) {
+
+    var outputState  = vni.outputState();
+    if ( _.isUndefined( errorFlag ) ) { 
+        outputState.error = ! _.isUndefined( vni.errorState().data );
+    } else { 
+        outputState.error = errorFlag;
+    }
+
+    if ( outputState.error ) { 
+         // If we have an error, we should forward the output so downstream nodes know about it
+         outputState.lm = createLm();
+    }
+
+    vni.outputState( outputState );
+}
+
+/** 
+ * Checks if the state was changed since the last state snapshot & if so, 
+ * updates the LM to ensure the change propagates.
+ *
+ * @param stateFacade a getter/setterfunction for the current state
+ * @param lastState a saved clone of the last state 
+ *
+ * @return returns true if state was updated, false if not.
+ */
+function stateChange( stateFacade, lastState ) { 
+
+    var currentState = stateFacade();
+
+    // If got new data or an error state
+    var lastData = (lastState &&  _.isObject( lastState )  ) ? lastState.data : lastState;
+    if ( currentState.data  !== lastData || currentState.error ) {    
+
+        currentState.lm = createLm();
+        stateFacade( currentState );
+        return true;
+    }
+
+    return false;
+}
+
+/*******************************************************************************/
+/* Support for default shouldRunUpdater policy                                 */
+/* Default policy checks to be sure there is data on all attached input ports. */
+/*******************************************************************************/                 
+
+/** 
+ * Retrieve all attached input ports for the noflo component associated with this
+ * node instance. 
+ *
+ * @param node instance context
+ *
+ * @return an array with the attached noflo port objects
+ */
+function attachedInputPorts( node ) { 
+
+    return _.filter( node.inPorts, 
+                     function( port) { return  port.listAttached().length > 0; });
+}
+
+/**
+ * returns true if there is data on all attached input ports; false if not
+ * 
+ * @param vni 
+ */
+function haveAllInputs( vni ) {
+
+    var attachedInPorts = attachedInputPorts( vni.node );
+
+    // If we have undefined data on any port, return false; if we 
+    // have data on all ports, return true
+    return ! _.some( attachedInPorts, function( port ) { 
+
+        var states = vni.inputStates( port.name );
+            
+        return ( _.isUndefined( states ) || 
+               ( _.isArray( states ) &&  _.some( states, _.isUndefined )));
+
+    });
 }
 
 /** 
@@ -135,30 +288,7 @@ function handlePortOutput( port, lastLm, state ) {
  * @return true if we have received data on all attached input port edges
  */
 function shouldRunUpdater( vni ) { 
-
-    var attachedInPorts = attachedInputPorts.call( vni.node );
-
-    // Check if we have input on all input ports and their edges
-    for ( var i=0, max=attachedInPorts.length; i < max; i++ ) { 
-
-        var port = attachedInPorts[i];
-        var states = vni.inputStates( port.name );
-            
-        // figure out how many input states we have received
-        var numberOfStates = 0;
-        if ( _.isArray( states ) ) { 
-           // filter out any undefined states and then count what we really have
-           var flattenedStates = _.filter( states, 
-                                           function(state) { return ! _.isUndefined(state); });
-           numberOfStates = flattenedStates.length;
-        } else if ( _.isObject( states ) ) {
-            numberOfStates = 1;
-        }
-
-        if (( numberOfStates === 0 ) || ( port.listAttached().length != numberOfStates )) {
-            return false;
-        }
-    }
- 
-    return true;
+    // TODO: May add different policies in the future.  This is why we 
+    // do NOT call the function below directly
+    return haveAllInputs( vni );
 }
--- a/src/framework-ondata.js
+++ b/src/framework-ondata.js
@@ -43,12 +43,13 @@ module.exports = function( payload, socketIndex ) {
              throw Error( 'No wrapper fRunUpdater function found!  Cannot run updater.' );
 
          } else { 
+             // TODO: revisit this logic later, per rdf-pipeline/noflo-rdf-pipeline#35
 
              // save data used to see if we have a different state after running updater
              var lastErrorState = _.clone( vni.errorState() );  // Shallow clone 
              var lastOutputLm = vni.outputState().lm;
  
-             clearStateData( _.partial( vni.errorState ) ); // clear last error state
+             clearStateData( vni.errorState ); // clear last error state
 
              // Save wrapper to make it accessible from the Promise
              var wrapper = this.nodeInstance.wrapper;
@@ -61,13 +62,11 @@ module.exports = function( payload, socketIndex ) {
              }).then( function() { 
                  // fRunUpdater/Updater success path
              
-
                  // Returned OK, but updater could have set an error - check for that
-                 if ( stateChange( _.partial( vni.errorState ), lastErrorState ) ) {
-                    lastOutputLm = undefined;  // got a new error error so we will force sending output
+                 if ( stateChange( vni.errorState, lastErrorState ) ) {
+                    setOutputErrorFlag( vni );   
+                    lastOutputLm = undefined;  // got a new error so we will force sending output
                  }
-
-                 setOutputErrorFlag( vni );   // set error flag to true if new error, false if not
 
                  // If not new output or error state data, send it downstream or log it to the console
                  handleOutput( outputPorts.output, lastOutputLm, vni.outputState() );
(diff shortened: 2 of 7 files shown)


src/framework-ondata.js
```diff
@@ -1,18 +1,129 @@
...
  */
 module.exports = function( payload, socketIndex ) {
 
+     var portName = this.name;
+
+     var vnid = payload.vnid || '';
+     var vni = this.nodeInstance.vni(vnid);
+
+     // Set the input states
+     vni.inputStates( portName, socketIndex, stateFactory( vnid, payload ) );
```
dbooth-boston: These lines should be checking to see if the payload is _already_ a state object (as evidenced by the presence of the payload.vnid property, even if its value is the empty string).  A new state object should only be created if the payload is not already a state object.  The reason for this is so that in the normal case, with RDF pipeline nodes connected to RDF pipeline nodes, downstream nodes directly reference the states of upstream nodes, without needing to recreate states.  So lines 27-31 should be:

<pre>
if (!('vnid' in payload)) {
  // Payload is from IIP or a vanilla NoFlo component.
  // Turn it into a state object:
  payload = stateFactory('', payload);
}
var vnid = payload.vnid || '';
var vni = this.nodeInstance.vni(vnid);

// Set the input states
vni.inputStates( portName, socketIndex, payload );
</pre>

src/framework-ondata.js
```diff
@@ -1,18 +1,129 @@
...
+
+     // check if should run updater & run if necessary
+     if ( shouldRunUpdater( this.nodeInstance, vni ) ) {
```
dbooth-boston: Now that we have added the node instance to the vni, I think we can just make this:

<pre>
if ( shouldRunUpdater( vni ) ) {
</pre>

src/framework-ondata.js
```diff
@@ -1,18 +1,129 @@
...
+        var outputPorts = this.nodeInstance.outPorts;
+
+        if ( _.isUndefined( this.nodeInstance.wrapper.fRunUpdater ) || 
+             ! _.isFunction( this.nodeInstance.wrapper.fRunUpdater ) ) { 
+
+             // Don't have a wrapper fRunUpdater 
+             throw Error( 'No wrapper fRunUpdater function found!  Cannot run updater.' );
+
+        } else { 
+            var promise = this.nodeInstance.wrapper.fRunUpdater.call( vni );
```
dbooth-boston: Would it make sense for the fRunUpdater to be called from here _inside_ a promise, so that if it throws an error then it will be caught and the reject path will be taken, instead of requiring every fRunUpdater to catch errors?

src/framework-ondata.js
```diff
@@ -1,18 +1,129 @@
...
+            Promise.resolve( promise ).then( function( data ) {
```
dbooth-boston: 'function(data)' can just be 'function()', because fRunUpdater does not need to return anything.  The framework will look at the output.lm and error.lm to detect state changes.

Also, a promise noobie question: Is Promise.resolve(promise).then(...) pretty much equivalent to promise.then(...) , or is the intervening 'resolve' important?

src/framework-ondata.js
glenna-m: this is being executed as a promise.  If an error is thrown, the promise will be rejected and will execute that error path.  I will ensure we have a test to demonstrate that.

src/framework-ondata.js
glenna-m: Resolve is what indicates that the promise has successfully returned.  Then is about what you want to do after resolve.

src/framework-ondata.js
glenna-m: will do.

src/framework-ondata.js
glenna-m: Noted.  Will address in next check in.

src/framework-ondata.js
```diff
@@ -1,18 +1,129 @@
...
+                    outputPorts.output.disconnect();
+                }
+            }, function( error ) { 
+    
+                // Set error state and send it on over the error port
+                var errorState = stateFactory( vnid, error );
+                vni.errorState( errorState );
+
+                outputPorts.error.output.send( errorState );
+                outputPorts.error.disconnect();
```
dbooth-boston: Can we get a sendPayload (abstraction method) added to our noflo access layer, so that we do not have to separately call send and disconnect here?   That sequence of calling send and disconnect is noflo-specific.

src/framework-ondata.js
```diff
@@ -1,18 +1,129 @@
...
+             // Don't have a wrapper fRunUpdater 
+             throw Error( 'No wrapper fRunUpdater function found!  Cannot run updater.' );
+
+        } else { 
+            var promise = this.nodeInstance.wrapper.fRunUpdater.call( vni );
+            Promise.resolve( promise ).then( function( data ) { 
+
+                // Set output state and send it on over the output port
+                if (data) {
+                    var outputState = stateFactory( vnid, data );
```
dbooth-boston: The framework should not create a new output state here.  It should already exist, as created by the updater or fRunUpdater.

glenna-m: It might be helpful to document which structures you expect the updater to manipulate.  I was thinking the states were wrapper/framework specific, but it sounds like you are thinking otherwise.

src/framework-ondata.js
```diff
@@ -1,18 +1,129 @@
...
+        // Save vars we will need in the promise where the context is different
+        var outputPorts = this.nodeInstance.outPorts;
+
+        if ( _.isUndefined( this.nodeInstance.wrapper.fRunUpdater ) || 
+             ! _.isFunction( this.nodeInstance.wrapper.fRunUpdater ) ) { 
+
+             // Don't have a wrapper fRunUpdater 
+             throw Error( 'No wrapper fRunUpdater function found!  Cannot run updater.' );
+
+        } else {
```
dbooth-boston: In general the logic for this section should be something like the following.   But I have written it in pseudo-code because I am not yet comfortable enough with promises to know exactly how it should be written.

<pre>
// Remember the old LMs, in order to detect changes make by the updater or fRunUpdater:
var oldOutputLm = vni.output.lm;
var oldErrorLm = vni.error.lm;

Run fRunUpdater, which returns nothing.

If the fRunUpdater threw an error (error), then:
  // Ensure errorState exists:
  if (!errorState) {
      errorState = stateFactory(vnid, error);
      vni.errorState(errorState);
  }
  // Make the error state change noticeable below:
  errorState.lm = createLm();
  errorState.data = error;

// In either case, continue.

// Update previousLms (if necessary), before sending events
if (vni.output.lm !== oldOutputLm) { 
  // TODO: vni.output.setPreviousLmsFromInputStates(vni);
}
if (vni.error.lm !== oldErrorLm) { 
  // TODO: vni.error.setPreviousLmsFromInputStates(vni);
}

// Send output event?
if (vni.output.lm !== oldOutputLm) { 
  outputPorts.output.sendPayload(outputState);
}

// Send error event?
if (vni.error.lm !== oldErrorLm) { 
  outputPorts.error.sendPayload(errorState);
}

// Log error event?
if (vni.error.lm !== oldErrorLm) { 
  if (vni.error.data) {
    logError(vni, vni.error.data);
  }
}
</pre>

src/framework-ondata.js
```diff
@@ -1,18 +1,129 @@
...
+ * This is simply a default updater policy - we may have other policies in the future.
+ * 
+ * @param node node instance of the RDF pipeline component
+ * @param vni virtual node instance whose input states are to be checked.
+ *
+ * @return true if we have received data on all required input port edges
+ */
+function shouldRunUpdater( node, vni ) { 
+
+    var requiredInPorts = requiredInputPorts.call( node );
```
dbooth-boston: This should be checking the _attached_ ports -- not the required ports.  The isRequired property only affects the configuration-time requirements -- not the runtime requirements.   Every attached port _must_ have data before potentially running the updater, required or not.   But it is a configuration-time error to run with a required port unattached.

src/framework-ondata.js
glenna-m: I think this sequence: 
  // Make the error state change noticeable below:
  errorState.lm = createLm();
  errorState.data = error;
defeats the purpose of using a facade.  Our main interface for doing this with the facade is imply set a new state e.g., vni.errorState( stateFactory(vnid, error) );  Any concerns with that?

dbooth-boston: Yes.  The data and lm properties of the state need to be individually settable.  The state object itself should only be created once, when the VNI is created.  Thereafter the same state object should always be used, with only the LM and data properties being changed.

dbooth-boston: Actually, come to think of it, both the output and error state objects should be created at VNI creation, with undefined data and lm properties -- not only the error object.  This is because the wrapper API does not need to include a state constructor -- only a destructor -- because state construction can be done by the wrapper's fRunUpdater.  (If needed, the fRunUpdater can see that the data property is undefined and do whatever it needs to initialize the state.)  Therefore it is fine for the framework to create the output state object (with undefined data and lm properties) at VNI creation.

glenna-m: undefined will be problematic for the lm.  Existing code expected an undefined lm means we should generate a new one.  I can override that by setting it after it returns, but that feels kludgy.  Do we need some kind of flag or other indicator?  Thoughts?

src/create-state.js
```diff
@@ -23,7 +23,14 @@ module.exports = function( vnid, data, lm ) {
            throw new Error("Unable to create state because no vnid was provided");
        }
 
-       var stateLm = lm || createLm();
+       var stateLm;
+       if ( _.isUndefined( data ) && _.isUndefined( lm ) ) { 
+           // If both data & lm are undefined, we are initializing - accept an undefined lm
```
dbooth-boston, line 28: good

src/create-state.js
```diff
@@ -23,7 +23,14 @@ module.exports = function( vnid, data, lm ) {
...
+           stateLm = lm;
```
jamesrdf: lm is undefined at this point

src/framework-ondata.js
dbooth-boston: The only case when the lm should be undefined is when the state object is first created (during VNI creation) but the updater has not yet been run.  The change that you made to create-state.js
 now addresses this case, so as far as I know nothing more is needed.  Do you see a place that would still be a problem?

src/create-state.js
catch-point: It might be clearer to the read if this line was just return {vnid: vnid};

src/framework-ondata.js
```diff
@@ -1,18 +1,164 @@
...
+     var inputState = (payload.vnid) ? payload : stateFactory( vnid, payload );
+     vni.inputStates( portName, socketIndex, inputState );
+
+     // check if should run updater & run if necessary
+     if ( shouldRunUpdater( vni ) ) { 
+          
+        // Save vars we will need in the promise where the context is different
+        var outputPorts = this.nodeInstance.outPorts;
+
+        if ( _.isUndefined( this.nodeInstance.wrapper.fRunUpdater ) ||
```
catch-point: If fRunUpdater is undefined it is also not a function, no need to check for both separately. However, it might be worth checking to see if the wrapper is present.

src/framework-ondata.js
```diff
@@ -1,18 +1,164 @@
...
+                    // TODO: vni.error.setPreviousLmsFromInputStates(vni);
+                    errorState.lm = createLm();
+                    errorState.data = rejected;
+                    vni.errorState( errorState );
+                    handlePortOutput( outputPorts.error, lastErrorLm, vni.errorState() );
+                 } 
+
+            }).catch( function() { 
+               console.error( "framework-ondata unable to process fRunUpdater results!" );
+               reject();
```
catch-point: reject is not defined anywhere. What do you expect this to do?

src/framework-ondata.js
```diff
@@ -1,18 +1,164 @@
...
+    
+                console.error( rejected );
+
+                // fRunUpdater failed - is this a new error? 
+                var errorState = vni.errorState();
+
+                if ( lastErrorLm !== errorState.lm  ) { 
+
+                    // fRunUpdater or updater recorded a new lm (i.e., new error) - send it on
+                    handlePortOutput( outputPorts.error, lastErrorLm, errorState );
```
catch-point: If fRunUpdater or updater recorded a new lm (i.e., new error) _and_ throw a rejected error, they are likely different and the system should not swallow the second error, but pass it along as well.

src/framework-ondata.js
```diff
@@ -1,18 +1,164 @@
...
+    for ( var i=0, max=attachedInPorts.length; i < max; i++ ) { 
+
+        var port = attachedInPorts[i];
+        var states = vni.inputStates( port.name );
+            
+        // figure out how many input states we have received
+        var numberOfStates = 0;
+        if ( _.isArray( states ) ) { 
+           // filter out any undefined states and then count what we really have
+           var flattenedStates = _.filter( states,
```
catch-point: _.compact does this already, no need to use an inline function.

src/framework-ondata.js
```diff
@@ -1,18 +1,164 @@
...
+     vni.inputStates( portName, socketIndex, inputState );
+
+     // check if should run updater & run if necessary
+     if ( shouldRunUpdater( vni ) ) { 
+          
+        // Save vars we will need in the promise where the context is different
+        var outputPorts = this.nodeInstance.outPorts;
+
+        if ( _.isUndefined( this.nodeInstance.wrapper.fRunUpdater ) || 
+             ! _.isFunction( this.nodeInstance.wrapper.fRunUpdater ) ) {
```
dbooth-boston: Come to think of it, let's separate this into two different cases, with both cases allowing the pipeline to continue execution:
1. If there is no updater then write a warning to the error state (and log it to the console) just as if an updater had run and set error.data to that warning and updated error.lm .  
2. If there is an fRunUpdater but it is not a function, then write an error message to the error state (setting error.data and error.lm) and also log it to the console.

If we take this approach then I think it will be friendlier to users, because the misconfiguration of one node will not stop the whole pipeline from running.

test/framework-ondata-mocha.js
```diff
@@ -0,0 +1,526 @@
...
+
+            }).then( function( done ) { 
+               // Verify we got the output state we expect
+               done.vnid.should.equal( '' ); 
+               done.data.should.equal( executedFRunUpdater ); 
+
+               // Cleanup 
+               test.detachAllInputSockets( node );
+            }, function( fail ) { 
+               assert.isNotOk( fail );
```
catch-point, line 165: Since the test is returning a promise, there is no need to assert that it didn't fail. The test will fail if the promise is rejected.

test/framework-ondata-mocha.js
```diff
@@ -0,0 +1,526 @@
...
+                                  required: true
+                              }
+                          }
+                        }, 
+                        wrapper )
+            );
+
+            var logBuffer = '';
+            // Hide expected console error message from test output
+            sinon.stub( console, 'error', function (message) {
```
catch-point, line 616: While this is sometimes handy, it is not needed for this test and should be removed.

src/framework-ondata.js
```diff
@@ -1,18 +1,164 @@
...
+                    // Same LM as before we called fRunUpdater, but new error data was detected
+                    // TODO: vni.error.setPreviousLmsFromInputStates(vni);
+                    errorState.lm = createLm();
+                    errorState.data = rejected;
+                    vni.errorState( errorState );
+                    handlePortOutput( outputPorts.error, lastErrorLm, vni.errorState() );
+                 } 
+
+            }).catch( function() { 
+               console.error( "framework-ondata unable to process fRunUpdater results!" );
```
catch-point: This log message should also log the actual error message as well (it will be passed as a parameter this this catch function).

src/create-state.js
dbooth-boston: I would rather have the data and lm properties there, but with undefined values.  But I guess it could be clearer if written as:

<pre>
stateLm = undefined;
</pre>

src/framework-ondata.js
```diff
@@ -1,18 +1,164 @@
...
+                resolve( wrapper.fRunUpdater.call( vni ) );
+
+            }).then( function() { 
+                // If not new output or error state data, send it downstream or log it to the console
+                handlePortOutput( outputPorts.output, lastOutputLm, vni.outputState() );
+                handlePortOutput( outputPorts.error,  lastErrorLm, vni.errorState() );
+
+            }, function( rejected ) { 
+    
+                console.error( rejected );
```
catch-point: Only log the error if there are no sockets attached to the error port. Otherwise, the error could be logged multiple times by each downstream node and make the log file confusing to understand.

test/framework-ondata-mocha.js
```diff
@@ -0,0 +1,526 @@
...
+
+        it( "should catch exception when fRunUpdater post processing fails.", function() {
+
+            var inData = "A bit of input data";
+            var executedFRunUpdater = "fRunUpdater success!!!";
+    
+            // Define the fRunUpdater that framework should invoke
+            var fRunUpdater = function() { 
+                // update state
+                this.outputState( undefined );
```
catch-point: This may not always produce an Error at the same place it the code. What about changing the outputState function to throw an error when it is called from within the framework-ondata?

src/framework-ondata.js
```diff
@@ -1,18 +1,164 @@
...
+            var lastErrorLm = vni.errorState().lm;
+            var wrapper = this.nodeInstance.wrapper;
+
+            new Promise(function( resolve ) { 
+                resolve( wrapper.fRunUpdater.call( vni ) );
+
+            }).then( function() { 
+                // If not new output or error state data, send it downstream or log it to the console
+                handlePortOutput( outputPorts.output, lastOutputLm, vni.outputState() );
+                handlePortOutput( outputPorts.error,  lastErrorLm, vni.errorState() );
```
dbooth-boston: This error should also be sent to the console _if_ it is non-empty.

src/framework-ondata.js
dbooth-boston: Good point.  And it raises the issue of what assumptions we should make about the content of error.data , and how we can distinguish between warnings and errors.  We need to distinguish because when an upstream node is in an error state then the downstream node's updater must not fire.  We should discuss how to do this.  Maybe consider a node to be in an error state if error.data is a string and contains the word "error", or if error.data is an object of type Error?

src/framework-ondata.js
```diff
@@ -1,18 +1,164 @@
...
+
+            }, function( rejected ) { 
+    
+                console.error( rejected );
+
+                // fRunUpdater failed - is this a new error? 
+                var errorState = vni.errorState();
+
+                if ( lastErrorLm !== errorState.lm  ) {
```
dbooth-boston: This test is already being done in handlePortOutput.  I think it should be done in only one place or the other, but not both.

src/framework-ondata.js
```diff
@@ -1,18 +1,164 @@
...
+                    handlePortOutput( outputPorts.error, lastErrorLm, errorState );
+
+                 } else if ( rejected !== errorState.data ) { 
+
+                    // Same LM as before we called fRunUpdater, but new error data was detected
+                    // TODO: vni.error.setPreviousLmsFromInputStates(vni);
+                    errorState.lm = createLm();
+                    errorState.data = rejected;
+                    vni.errorState( errorState );
+                    handlePortOutput( outputPorts.error, lastErrorLm, vni.errorState() );
```
dbooth-boston: s/vni.errorState()/errorState/

src/framework-ondata.js
```diff
@@ -1,18 +1,164 @@
...
+
+/** 
+ * Retrieve all attached input ports for the noflo component associated with this
+ * node instance. 
+ *
+ * @this node instance context
+ *
+ * @return an array with the attached noflo port objects
+ */
+function attachedInputPorts() {
```
dbooth-boston: Suggest moving this to the node facade.

src/framework-ondata.js
glenna-m: Will remove it .  I was hoping it would give me an immediate promise reject.  It does not fail so it is present in emca script 6, but it did not come back immediately either.

src/framework-ondata.js
```diff
@@ -1,18 +1,294 @@
...
+             // Don't have a wrapper fRunUpdater 
+             throw Error( 'No wrapper fRunUpdater function found!  Cannot run updater.' );
+
+         } else { 
+
+             // save data used to see if we have a different state after running updater
+             var lastErrorState = _.clone( vni.errorState() );  // Shallow clone 
+             var lastOutputLm = vni.outputState().lm;
+ 
+             clearStateData( _.partial( vni.errorState ) ); // clear last error state
```
dbooth-boston: AFAICT _.partial does not bind 'this' and therefore does nothing when called with a single function argument, so the above could be written as:

<pre>
clearStateData( vni.errorState );
</pre>

Ditto in several other places.

src/framework-ondata.js
```diff
@@ -1,18 +1,294 @@
...
+    var outputState  = vni.outputState();
+    if ( _.isUndefined( errorFlag ) ) { 
+        outputState.error = ! _.isUndefined( vni.errorState().data );
+    } else { 
+        outputState.error = errorFlag;
+    }
+
+    if ( outputState.error ) { 
+         // If we have an error, we should forward the output so downstream nodes know about it
+         outputState.lm = createLm();
```
dbooth-boston: No need to change this now, but I've logged an issue to change this sometime in the future.  The downstream nodes should be notified, but the outputState.lm should not be changed, for the reason explained more fully here:
https://github.com/rdf-pipeline/noflo-rdf-pipeline/issues/33

Maybe just put a TODO comment here for the moment, referencing issue 33.

src/framework-ondata.js
```diff
@@ -1,18 +1,294 @@
...
+ */
+function stateChange( stateFacade, lastState ) { 
+
+    var currentState = stateFacade();
+
+    // If got new data or an error state
+    var lastData = (lastState &&  _.isObject( lastState )  ) ? lastState.data : lastState;
+    if ( currentState.data  !== lastData || currentState.error ) {    
+
+        currentState.lm = createLm();
```
dbooth-boston: Add a TODO comment referencing https://github.com/rdf-pipeline/noflo-rdf-pipeline/issues/33

src/framework-ondata.js
```diff
@@ -1,18 +1,294 @@
...
+
+    var attachedInPorts = attachedInputPorts( vni.node );
+
+    // If we have undefined data on any port, return false; if we 
+    // have data on all ports, return true
+    return ! _.some( attachedInPorts, function( port ) { 
+
+        var states = vni.inputStates( port.name );
+            
+        return ( _.isUndefined( states ) ||
```
dbooth-boston: No need to change this now, but please add a TODO comment referencing https://github.com/rdf-pipeline/noflo-rdf-pipeline/issues/34 about also checking state.error .

src/framework-ondata.js
```diff
@@ -1,18 +1,294 @@
...
+         var outputPorts = this.nodeInstance.outPorts;
+
+         if ( _.isUndefined( this.nodeInstance.wrapper.fRunUpdater ) || 
+             ! _.isFunction( this.nodeInstance.wrapper.fRunUpdater ) ) { 
+
+             // Don't have a wrapper fRunUpdater 
+             throw Error( 'No wrapper fRunUpdater function found!  Cannot run updater.' );
+
+         } else { 
+
```
dbooth-boston, line 46: Please add a TODO comment to revisit this logic later, per https://github.com/rdf-pipeline/noflo-rdf-pipeline/issues/35

test/framework-ondata-mocha.js
```diff
@@ -0,0 +1,724 @@
...
+describe("framework-ondata", function() {
+
+    it( "should exist as a function", function() {
+      framework_ondata.should.exist;
+      framework_ondata.should.be.a('function');
+    });
+
+    describe("#ondata", function() {
+
+        it( "should throw error if no wrapper fRunUpdater configured", function() {
```
dbooth-boston, line 31: good

test/framework-ondata-mocha.js
```diff
@@ -0,0 +1,724 @@
...
+            // We use a promise here because this section is asynchronous
+            return new Promise( function(done, fail) {
+                // Send some data to the input port
+                test.onOutPortData(node, 'output', done);
+                test.onOutPortData(node, 'error', fail);
+                test.sendData(node, 'input', inData);
+            }).should.be.rejectedWith('No wrapper fRunUpdater function found!  Cannot run updater.');
+        });
+
+        it( "should manage single port node vni state, fRunUpdater invocation, & output state", function() {
```
dbooth-boston, line 58: good

test/framework-ondata-mocha.js
```diff
@@ -0,0 +1,724 @@
...
+               done.vnid.should.equal( '' ); 
+               done.data.should.equal( executedFRunUpdater ); 
+
+            }, function( fail ) { 
+               // Failure - something is not right
+               assert.isNotOk( fail );
+            });
+        });
+
+        it( "should handle two input/output state changes", function() {
```
dbooth-boston, line 107: good

test/framework-ondata-mocha.js
```diff
@@ -0,0 +1,724 @@
...
+            var errors = [ "Error 1", "Error 2" ];
+            var outData = [ "Data 1", "Data 2" ];
+
+            // Define the fRunUpdater that framework should invoke
+            var fRunUpdater = function() { 
+
+                // update error state
+                var errState = this.errorState();
+                errState.data = errors[count];
+                errState.lm = createLm();
```
dbooth-boston, line 183: Add a TODO comment referencing https://github.com/rdf-pipeline/noflo-rdf-pipeline/issues/37

test/framework-ondata-mocha.js
```diff
@@ -0,0 +1,724 @@
...
+               }, function( fail2 ) { 
+                   assert.isNotOk( fail2 );
+               });
+
+            }, function( fail ) { 
+               assert.isNotOk( fail );
+            });
+        });
+
+        it( "should handle two updater set error state changes", function() {
```
dbooth-boston, line 169: good

test/framework-ondata-mocha.js
```diff
@@ -0,0 +1,724 @@
...
+                  fail2.vnid.should.equal( '' ); 
+                  fail2.data.should.equal( errors[1] ); 
+               });
+
+            }, function( fail ) { 
+               assert.isNotOk( fail );
+            });
+        });
+
+        it( "should handle two updater failures", function() {
```
dbooth-boston, line 237: good

test/framework-ondata-mocha.js
```diff
@@ -0,0 +1,724 @@
...
+                                                         required: true
+                                                     }
+                                                 }
+                                               }, 
+                                               wrapper )
+                                         }
+            }).then( function( network ) {
+                return new Promise( function(done, fail) { 
+
+                    // True noflo component - not facade
```
dbooth-boston, line 344: helpful comment.  thanks!

test/framework-ondata-mocha.js
```diff
@@ -0,0 +1,724 @@
...
+                    var node = network.processes.node3.component;
+
+                    test.onOutPortData(node, 'output', done);
+                    test.onOutPortData(node, 'error', fail);
+
+                    network.graph.addEdge( 'node1', 'out', 'node3', 'input1' );
```
dbooth-boston, line 350: This is much nicer than the low level noflo calls we had before.

test/framework-ondata-mocha.js
```diff
@@ -0,0 +1,724 @@
...
+
+                    network.graph.addInitial( input1, 'node1', 'in' );
+                    network.graph.addInitial( input2, 'node2', 'in' );
+
+                }).then( function( done ) { 
+
+                   // Stop network & verify we got the output state we expect
+                   network.stop();
+                   done.vnid.should.equal( '' ); 
+                   done.data.should.equal( executedFRunUpdater );
```
dbooth-boston, line 361: excellent

test/framework-ondata-mocha.js
```diff
@@ -0,0 +1,724 @@
...
+                   done.data.should.equal( executedFRunUpdater ); 
+
+                }, function( fail ) { 
+                   network.stop();
+                   assert.isNotOk( fail );
+                });
+            });
+        });
+
+        it( "should invoke fRunUpdater without waiting for unattached port data", function() {
```
dbooth-boston, line 370: good

test/framework-ondata-mocha.js
```diff
@@ -0,0 +1,724 @@
...
+                   done.data.should.equal( executedFRunUpdater ); 
+    
+                }, function( fail ) { 
+                   network.stop();
+                   assert.isNotOk( fail );
+                });
+            });
+        });
+
+        it( "should not invoke fRunUpdater with missing attached port data", function() {
```
dbooth-boston, line 445: good

test/framework-ondata-mocha.js
```diff
@@ -0,0 +1,724 @@
...
+                   network.stop();
+
+                }, function( fail ) { 
+                   network.stop();
+                   assert.isNotOk( fail );
+                });
+            });
+        });
+
+        it("should call fRunUpdater on addressable ports", function() {
```
dbooth-boston, line 512: good

dbooth-boston:
LGTM.  

Thanks for getting this out.  I noted a few things that we will want to revisit in the future, but let's not let them hold this up.

test/framework-ondata-mocha.js
glenna-m, line 616: I have it here because the error messages we are causing will print to the console and be pretty confusing to someone who does not know these tests.  They will wonder if there is an error here or not.  Let me show you what that would look like before we remove it.  It's not pretty.

dbooth-boston:
LGTM

src/framework-ondata.js
```diff
@@ -1,18 +1,298 @@
...
+     var inputState = (payload.vnid) ? payload : stateFactory( vnid, payload );
+     vni.inputStates( portName, socketIndex, inputState );
+
+     if ( shouldRunUpdater( vni ) ) { 
+
+         // Save vars we will need in the promise where the context is different
+         var outputPorts = this.nodeInstance.outPorts;
+
+         if ( _.isUndefined( this.nodeInstance.wrapper.fRunUpdater ) ||
```
catch-point: checking for undefined is redundant here, since the next line checks if it is a function

src/framework-ondata.js
```diff
@@ -1,18 +1,298 @@
...
+ 
+             clearStateData( vni.errorState ); // clear last error state
+
+             // Save wrapper to make it accessible from the Promise
+             var wrapper = this.nodeInstance.wrapper;
+
+             new Promise(function( resolve ) { 
+ 
+                 // Execute fRunUpdater which will also execute the updater
+                 resolve( wrapper.fRunUpdater.call( vni ) );
```
catch-point: Maybe fRunUpdater should take vni as a parameter.

src/framework-ondata.js
```diff
@@ -1,18 +1,298 @@
...
+             }, function( rejected ) { 
+                 // fRunUpdater/updater failed 
+                 if ( isInitState( vni.errorState ) ) { 
+                     // fRunUpdater/updater has not already set an error state - use the rejected info
+                     changeStateData( vni.errorState, rejected );
+                 }
+
+                 setOutputErrorFlag( vni, true ); 
+
+                 stateChange( vni.outputState, lastOutputLm );
```
catch-point: Since stateChange will not use the second parameter (if it's not an object) better to use a static value here, like false or undefined, or better yet don't include a second parameter.

src/framework-ondata.js
```diff
@@ -1,18 +1,298 @@
...
+                     lastErrorState = _.clone( vni.errorState() );   
+                     changeStateData( vni.errorState, rejected );
+                     if ( changeStateData( vni.errorState, lastErrorState ) ) { 
+                         lastErrorState.lm = undefined; 
+                     } 
+                     handleError( outputPorts.error, lastErrorState.lm, vni.errorState() );
+                 }
+
+             }).catch( function( e ) { 
+                 // fRunUpdater or Updater threw an error
```
catch-point: This would not catch an error from fRunUpdate or Updater (directly) only errors from the function starting on line 75. Change this comment to something like // Could not handle output or error result

src/framework-ondata.js
```diff
@@ -1,18 +1,298 @@
...
+                 console.error( "framework-ondata unable to process fRunUpdater results!" );
+                 var err = ( e.stack ) ? e.stack : e;
```
catch-point, line 107: This err value might be useful for debugging, please log it.

src/framework-ondata.js
```diff
@@ -1,18 +1,301 @@
...
+
+                 handleOutput( outputPorts.output, lastOutputLm, vni.outputState() );
+                 handleError( vni, outputPorts.error, lastErrorState.lm );
+                  
+
+                 // If we haven't already processed the rejected error, do it now
+                 if ( rejected !== vni.errorState().data ) { 
+                     lastErrorState = _.clone( vni.errorState() );   
+                     changeStateData( vni.errorState, rejected );
+                     if ( changeStateData( vni.errorState, lastErrorState ) ) {
```
catch-point: The function changeStateData has no return statement, so this condition would never be fulfilled.

src/framework-ondata.js
```diff
@@ -1,18 +1,301 @@
...
+
+/**
+ * change state data to the specified newData value
+ *
+ * @param stateFacade a getter/setterfunction for the current state
+ */ 
+function changeStateData( stateFacade, newData, newLm ) { 
+    var currentState = stateFacade();
+    currentState.data = newData;
+    if ( newLm ) {
```
catch-point: newLm is never used

test/framework-ondata-mocha.js
catch-point, line 616: fair enough

src/framework-ondata.js
```diff
@@ -1,18 +1,301 @@
...
+ * @param lastState a saved clone of the last state 
+ *
+ * @return returns true if state was updated, false if not.
+ */
+function stateChange( stateFacade, lastState ) { 
+
+    var currentState = stateFacade();
+
+    // If got new data or an error state
+    var lastData = (lastState &&  _.isObject( lastState )  ) ? lastState.data : lastState;
```
catch-point: _.isObject is sufficient here (no need to check lastState too)

src/framework-ondata.js
glenna-m: newLm is used to determine if we need to create a new one or just use the one we already have.

src/framework-ondata.js
```diff
@@ -1,18 +1,301 @@
...
+    if ( currentState.data  !== lastData ) { 
+
+        currentState.lm = createLm();
```
dbooth-boston: Only generate a new LM for error object -- not for output object.  Better to do it outside of this function, so that this function does not have a side effect.

catch-point:
LGTM

PR #59, opened 2016-03-10
Add chcs-patient-jsonld graph

catch-point:
A subgraph component that populates the patientid into an embedded URL template and parses the resulting JSON into an object. The authority is optional to allow local modification for testing and staging.


glenna-m:
LGTM

dbooth-boston:
LGTM

PR #64, opened 2016-03-24
Js wrapper

glenna-m:
This PR covers the following functionality: 
1. Javascript wrapper + behavioral tests
2. Iteration 5 behavioral tests and test component - this component is written in the new simplified style
3. HTML 2 object comparison component - this one is written in the older style where ports with settings like required & multiple inputs are specified.

I did break them into multiple commits at James's request to allow for easier cherry picking/merging into this work.

### DIFF
--- a/components/merge-patient-lab-iips.js
+++ b/components/merge-patient-lab-iips.js
@@ -14,12 +14,57 @@ var jswrapper = require('../src/javascript-wrapper');
  *
  * @param patient a JSON patient record to be processed
  * @param labwork a JSON lab record to be processed
- * @this component context
+ * @this vni context
  *
  * @return the combined patient/lab record
  */
-module.exports = jswrapper(function(patient, labwork) {
-    patient = _.isString(patient) ? JSON.parse(patient) : patient;
-    labwork = _.isString(labwork) ? JSON.parse(labwork) : labwork;
+module.exports = jswrapper( updater );
+
+/**
+ * Merges a patient and a labwork record into a single combined record.
+ * If no patient or labwork parameter is specified, the original parameter
+ * will be returned.
+ * 
+ * @this like all RDF pipeline updaters, this will usually be called on the 
+ *       vni context, however, this updater does not depend on that context. 
+ *
+ * @param patient a patient record in either json or string format
+ * @param labwork a lab record in either json or string format
+ * 
+ * @return the merged record
+ */
+function updater(patient, labwork) {
+
+    patient = jsObject( patient );
+    labwork = jsObject( labwork );
     return _.extend({}, patient, labwork);
-});
+}
+
+/**
+ * Converts whatever input parameter of any type (or even undefined) 
+ * into a javascript object 
+ * 
+ * @param parameter input parameter to be converted.  This could be
+ *        a string, already a javascript object, or even undefined.
+ *
+ * @return a javascript object 
+ */
+function jsObject( parameter ) { 
+
+    if ( _.isUndefined( parameter ) ) {  
+        return {}; 
+    } 
+
+    try { 
+        return  _.isString(parameter) ? JSON.parse(parameter) : parameter;
+    } catch ( e ) { 
+       var msg = "Unable to parse parameter "+parameter+": "+e;
+       console.error( msg );
+       throw Error( msg );
+    }
+} 
+
+// If running in test mode, mocha will execute internal unit testing
+if (process.env.NODE_ENV === 'test') {
+  module.exports._private = { updater: updater };
+}
--- a/components/merge-patient-lab-iips.js
+++ b/components/merge-patient-lab-iips.js
@@ -63,8 +63,3 @@ function jsObject( parameter ) {
        throw Error( msg );
     }
 } 
-
-// If running in test mode, mocha will execute internal unit testing
-if (process.env.NODE_ENV === 'test') {
-  module.exports._private = { updater: updater };
-}
--- a/src/javascript-wrapper.js
+++ b/src/javascript-wrapper.js
@@ -133,14 +133,9 @@ module.exports = function( nodeDefOrUpdater ) {
 					       }, {}));
     }
 
-    return factory( nodeDef,
-                    { 
-                      fRunUpdater: fRunUpdater.bind( this, 
-                                                     updater, 
-                                                     updaterArgs ) 
-                      // TODO: Add other Wrapper API's here
-                    } 
-           );
+    // TODO: Add additional wrapper functions to this object
+    var wrapper = {fRunUpdater: fRunUpdater.bind(this, updater, updaterArgs)};
+    return _.defaults(factory(nodeDef, wrapper), nodeDef);
 
 }; // module.exports
 
--- a/src/javascript-wrapper.js
+++ b/src/javascript-wrapper.js
@@ -1,0 +1,193 @@
+/**
+ * File: javascript-wrapper.js 
+ */
+var _ = require('underscore');
+
+var createLm = require('./create-lm');
+var createState = require('./create-state');
+var factory = require('./pipeline-component-factory');
+var util = require('util');
+
+/**
+ * RDF Pipeline javascript wrapper fRunUpdater API as documented here: 
+ *    https://github.com/rdf-pipeline/noflo-rdf-pipeline/wiki/Wrapper-API
+ * 
+ * @param updater component updater function
+ * @param updaterArgs the introspected arguments found for the component updater.
+ *        used to extract the input for the updater
+ * @param vni a virtual node instance
+ */
+var fRunUpdater = function( updater, updaterArgs, vni ) {
+
+    // Default to no updater parameters; if no arguments, we won't go through the map
+    var updaterParameters = [];
+
+    // Have one or more updater parameters - get them in the right order and pass them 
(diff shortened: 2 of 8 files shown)


components/merge-patient-lab-iips.js
```diff
@@ -14,12 +14,57 @@ var jswrapper = require('../src/javascript-wrapper');
...
+    } catch ( e ) { 
+       var msg = "Unable to parse parameter "+parameter+": "+e;
+       console.error( msg );
+       throw Error( msg );
+    }
+} 
+
+// If running in test mode, mocha will execute internal unit testing
+if (process.env.NODE_ENV === 'test') {
+  module.exports._private = { updater: updater };
```
catch-point: In a previous meeting we discussed (and demonstrated) changing pipeline-component-factory.js to expose the updater function for testing, without a test flag.

test/merge-patient-lab-iips-mocha.js
```diff
@@ -1,174 +1,263 @@
...
 var _ = require('underscore');
-var noflo = require('noflo');
 
 var compFactory = require('../components/merge-patient-lab-iips');
 
 var test = require('./common-test');
 var stubs = require('./common-stubs');
 
-describe('merge-patient-lab-iips', function() {
+describe( 'merge-patient-lab-iips', function() {
```
catch-point: It's no big deal to have, but since this modification is just adding the space I want to point out that these extra spaces are against our style guide AFAIK. "Do not add spaces after the open paren or before the close paren" See https://google.github.io/styleguide/cppguide.html#Function_Calls

test/merge-patient-lab-iips-mocha.js
```diff
@@ -1,174 +1,263 @@
...
 
-        it("should have outputState function", function() {
-            var node = test.createComponent(compFactory);
-            node.vni().outputState.should.exist;
-            node.vni().outputState.should.be.a('function');
-        });
+            it( "should initially return empty error state", function() {
+                var node = test.createComponent(compFactory);
+                var errorState = node.vni().errorState();
+                errorState.should.have.all.keys( 'vnid', 'data', 'error', 'lm' );
```
catch-point: I don't think we should distinguish between the keys being present and having an undefined value. I think this should be changed to errorState.should.be.an('object')

test/merge-patient-lab-iips-mocha.js
```diff
@@ -192,20 +281,6 @@ describe('merge-patient-lab-iips', function() {
             }).then(_.property('data')).should.become({id: '001',  glucose: '75',  date: '2012-02-01'});
         });
 
-        it( "should be able to call updater without sockets", function() {
```
catch-point, line 195: Why was this test removed?

test/merge-patient-lab-iips-mocha.js
```diff
@@ -231,5 +306,143 @@ describe('merge-patient-lab-iips', function() {
...
+                               '{"id 001, "glucose": "85", "date": "2013-02-01" }');
+
+             }).then(function(payload){
+
+                console.error.restore();
+                payload.should.exist;
+                payload.should.not.be.empty;
+                payload.should.have.all.keys( 'vnid', 'data', 'error', 'lm' );
+                payload.vnid.should.equal('');
+                payload.data.startsWith('Error: Unable to parse parameter').should.be.true;
```
catch-point, line 400: I don't think the output state data should contain the error message. I think it should be undefined in this case.

test/merge-patient-lab-iips-mocha.js
```diff
@@ -231,5 +306,143 @@ describe('merge-patient-lab-iips', function() {
...
+                console.error.restore();
+                assert.isNotOk( fail );
+             });
+        });
     });
+
+
+  // This introspect tests must be run in test mode with NODE_ENV=test 
+  // e.g., NODE_ENV=test mocha since we are testing a private function
+  describe('#updater - run with NODE_ENV=test to fully execute these tests', function() {
```
catch-point: I'd rather find another way to run this without an environment variable

test/merge-patient-lab-iips-mocha.js
glenna-m: Did not notice that in the guide.  I do it for readability, but recognize it can complicate searching.  I can stop that.

test/merge-patient-lab-iips-mocha.js
glenna-m: I disagree.  Object is  generic.  By having the keys, we are sure we actually have something that looks like an error object.  Let's discuss.

test/merge-patient-lab-iips-mocha.js
glenna-m, line 195: The description was misleading for what the test is really doing IMHO.

The functionality is still there under this test description:
it( "should have patient and labwork output state after input ports processing"

test/merge-patient-lab-iips-mocha.js
glenna-m: I think we have a difference in philosophy here.  

If we just trigger an updater run without the env var, we get something with a lot of other software in the stack - node, framework, javascript wrapper.   That is OK for true behavioural tests.

However, I think there is still some value in direct unit testing where you just hit the function you want to test.

src/javascript-wrapper.js
```diff
@@ -0,0 +1,188 @@
...
+         } 
+         resolve( results );
+
+    }).catch( function( e ) { 
+         var outputState = vni.outputState();
+         if ( _.isUndefined( outputState.error ) || ! outputState.error  ) { 
+             outputState.error = true;
+             outputState.data = e.toString();
+             outputState.lm = createLm();
+             vni.outputState( outputState );
```
dbooth-boston, line 56: Lines 56-58 should be setting errorState instead of outputState.  The outputState.data and outputState.lm should not be touched.

src/javascript-wrapper.js
```diff
@@ -0,0 +1,188 @@
...
+};
+
+/**
+ * This module provides a Javascript wrapper for updaters.  It is responsible for setting up the
+ * code to ensure updater callback, and then invoking the RDF pipeline factory to complete building
+ * the component and setting up the vni and state metadata.
+ *
+ * @param nodeDefOrUpdater 
+ *
+ * @this the context for this wrapper is the node instance.
```
dbooth-boston: AFAICT @this is not really used.  It is passed to fRunUpdater on line 137, but fRunUpdater does not use it.  I suggest deleting the @this comment on line 71 and changing line 137 to use _.partial instead of .bind .

src/javascript-wrapper.js
```diff
@@ -0,0 +1,188 @@
...
+       nodeDef.inPorts =  (_.isEmpty( undefinedInPortNames )) ? inPorts : 
+                               _.defaults( inPorts,
+                                           undefinedInPortNames.reduce(function(map, obj) {
+                                               map[obj] = { datatype: 'all' };
+                                               return map;
+					       }, {}));
+    }
+
+    // TODO: Add additional wrapper functions to this object
+    var wrapper = {fRunUpdater: fRunUpdater.bind(this, updater, updaterArgs)};
```
dbooth-boston: As commented above, I suggest eliminating 'this' (because it isn't used) and changing bind to _.partial:

<pre>
var wrapper = {fRunUpdater: _.partial(fRunUpdater, updater, updaterArgs)};
</pre>

src/javascript-wrapper.js
catch-point, line 56: I would expect that the error flag be set to true here (as it is) for the output state.

test/javascript-wrapper-mocha.js
```diff
@@ -0,0 +1,275 @@
...
+            }).should.eventually.contain('input');
+        });
+
+        it("should accept array of inPort names", function() {
+            return Promise.resolve({
+                inPorts:['testport']
+            }).then(jswrapper).then(commonTest.createComponent).then(function(component){
+                return _.keys(component.inPorts).sort();
+            }).should.eventually.become(['input', 'testport']);
+        });
```
dbooth-boston, line 55: nice, I like these tests

test/javascript-wrapper-mocha.js
```diff
@@ -0,0 +1,275 @@
...
+
+        it("should accept object with one inPort definition", function() {
+            return Promise.resolve({
+                inPorts: { 
+                    input: { datatype: 'string',
```
dbooth-boston: Suggestion: Use a different port name, so that it is distinguishable from the default case.

test/javascript-wrapper-mocha.js
```diff
@@ -0,0 +1,275 @@
...
+         return new Promise(function(callback){
+             handler = callback;
+             commonTest.sendData(node, 'input', "test input");
+         }).should.become( 'success' );
+    });
+  });
+
+  // This introspect tests must be run in test mode with NODE_ENV=test 
+  // e.g., NODE_ENV=test mocha since we are testing a private function
+  describe('#introspect - run with NODE_ENV=test to fully execute these tests', function() {
```
dbooth-boston, line 231: This will be updated to remove NODE_ENV?

dbooth-boston:
LGTM

catch-point:
LGTM

PR #65, opened 2016-03-24
Encode issue 35 as behaviour tests

catch-point:
See https://github.com/rdf-pipeline/noflo-rdf-pipeline/issues/35 for a description of the semantics

### DIFF
--- a/test/jsupdate-behaviour-mocha.js
+++ b/test/jsupdate-behaviour-mocha.js
@@ -267,7 +267,7 @@ describe("jsupdater-behaviour", function() {
             }).catch(fail);
         }).should.eventually.have.property('data', "Hello once");
     });
-    it("should fire updater if upstream updater did nothing, after error, after valid output", function() {
+    it("should produce output, but not fire updater if upstream updater did nothing, after error, after valid output", function() {
         return new Promise(function(done, fail){
             var count = 0;
             test.createNetwork({
@@ -279,7 +279,10 @@ describe("jsupdater-behaviour", function() {
                     }
                 }),
                 sut: jswrapper(function(input) {
-                    return "Hello " + input;
+                    switch(count) {
+                        case 1: return "Hello " + input;
+                        default: fail("Hello " + input);
+                    }
                 })
             }).then(function(network){
                 network.graph.addEdge('upstream', 'output', 'sut', 'input');
--- a/test/jsupdate-behaviour-mocha.js
+++ b/test/jsupdate-behaviour-mocha.js
@@ -221,6 +221,31 @@ describe("jsupdater-behaviour", function() {
             }).catch(fail);
         }).should.eventually.have.property('data', "Hello upstream");
     });
+    it("should not fire updater if upstream updater corrected error, but has no state", function() {
+        return new Promise(function(done, fail){
+            var count = 0;
+            test.createNetwork({
+                upstream: jswrapper(function(input) {
+                    switch(++count) {
+                        case 1: throw input;
+                        case 2: return;
+                    }
+                }),
+                sut: jswrapper(function(input) {
+                    fail("Hello " + input);
+                })
+            }).then(function(network){
+                network.graph.addEdge('upstream', 'output', 'sut', 'input');
+                return new Promise(function(adv) {
+                    test.onOutPortData(network.processes.upstream.component, 'error', adv);
+                    test.sendData(network.processes.upstream.component, 'input', "upstream");
+                }).then(function() {
+                    test.sendData(network.processes.upstream.component, 'input', "correction");
+                    setTimeout(done.bind(this, "nothing happened"), 100);
+                });
+            }).catch(fail);
+        }).should.become("nothing happened");
+    });
     it("should not fire updater if upstream updater set an error data", function() {
         return new Promise(function(done, fail){
             return test.createNetwork({
--- a/test/jsupdate-behaviour-mocha.js
+++ b/test/jsupdate-behaviour-mocha.js
@@ -71,6 +71,7 @@ describe("jsupdater-behaviour", function() {
                 once: jswrapper(function(input) {
                     switch(++count) {
                         case 1: return input;
+                        case 2: return;
                     }
                 }),
                 sut: jswrapper(function(input) {
@@ -197,6 +198,7 @@ describe("jsupdater-behaviour", function() {
                     switch(++count) {
                         case 1: return input;
                         case 2: throw input;
+                        case 3: return;
                     }
                 }),
                 sut: jswrapper(function(input) {
--- a/test/jsupdate-behaviour-mocha.js
+++ b/test/jsupdate-behaviour-mocha.js
@@ -155,6 +155,8 @@ describe("jsupdater-behaviour", function() {
             }).then(function(network){
                 network.graph.addEdge('once', 'output', 'sut', 'input');
                 return new Promise(function(adv) {
+                    // This is in a promise because we need to wait for the first event to finish propagating
+                    // before registering 'fail' on the output port a few lines below.
                     test.onOutPortData(network.processes.sut.component, 'output', adv);
                     test.sendData(network.processes.once.component, 'input', "once");
                 }).then(function() {
--- a/test/jsupdate-behaviour-mocha.js
+++ b/test/jsupdate-behaviour-mocha.js
@@ -13,7 +13,7 @@ var jswrapper = require('../src/javascript-wrapper');
  * Theses test are intented to verify the state management of the system.
  * @see https://github.com/rdf-pipeline/noflo-rdf-pipeline/issues/35
  */
-describe("framework-state", function() {
+describe("jsupdater-behaviour", function() {
     it("should fire updater when it has valid input states on all of its attached inputs", function() {
         return new Promise(function(done, fail){
             return test.createNetwork({
--- a/test/jsupdate-behaviour-mocha.js
+++ b/test/jsupdate-behaviour-mocha.js
@@ -82,6 +82,31 @@ describe("jsupdater-behaviour", function() {
             }).catch(fail);
         }).should.eventually.have.property('data', "Hello from node1 and from node2");
     });
+    it("should fire updater when it has valid input states on all of its attached sockets", function() {
+        return new Promise(function(done, fail){
+            test.createNetwork({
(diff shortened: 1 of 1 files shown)


test/jsupdate-behaviour-mocha.js
```diff
@@ -0,0 +1,156 @@
...
+                test.sendData(network.processes.node1.component, 'in', "from node1");
+                setTimeout(done.bind(this, "nothing happened"), 100);
+            });
+        }).should.become("nothing happened");
+    });
+    it("should not fire updater if upstream updater did not produce an output state", function() {
+        return new Promise(function(done, fail){
+            return test.createNetwork({
+                sink: jswrapper(function(input) {
+                    // do nothing
```
dbooth-boston, line 131: Note: This looks right, but I think that the framework will have to be modified to handle this case.

test/jsupdate-behaviour-mocha.js
```diff
@@ -0,0 +1,156 @@
...
+                test.sendData(network.processes.sink.component, 'input', "tongue");
+                setTimeout(done.bind(this, "nothing happened"), 100);
+            });
+        }).should.become("nothing happened");
+    });
+    it("should not fire updater if upstream updater set an error", function() {
+        return new Promise(function(done, fail){
+            return test.createNetwork({
+                broken: jswrapper(function(input) {
+                    this.errorState({data:input});
```
dbooth-boston: Should be:

<pre>
this.outputState().error = true;
</pre>

test/jsupdate-behaviour-mocha.js
```diff
@@ -45,12 +45,12 @@ describe("framework-state", function() {
                 network.graph.addEdge('node2', 'out', 'sut', 'input2');
                 test.sendData(network.processes.node1.component, 'in', "from node1");
                 setTimeout(done.bind(this, "nothing happened"), 100);
-            });
+            }).catch(fail);
         }).should.become("nothing happened");
```
glenna-m, line 125: Good pattern to follow.  I'll be adopting this.

test/jsupdate-behaviour-mocha.js
```diff
@@ -61,14 +136,40 @@ describe("framework-state", function() {
                 network.graph.addEdge('sink', 'output', 'sut', 'input');
                 test.sendData(network.processes.sink.component, 'input', "tongue");
                 setTimeout(done.bind(this, "nothing happened"), 100);
-            });
+            }).catch(fail);
+        }).should.become("nothing happened");
+    });
+    it("should not fire updater if upstream updater did not produce an output state after producing one", function() {
```
dbooth-boston: s/did not produce an output state after producing one/returns undefined after producing an output state/

test/jsupdate-behaviour-mocha.js
```diff
@@ -61,14 +136,40 @@ describe("framework-state", function() {
...
+                        case 2: return;
+                    }
+                }),
+                sut: jswrapper(function(input) {
+                    return "Hello " + input;
+                })
+            }).then(function(network){
+                network.graph.addEdge('once', 'output', 'sut', 'input');
+                return new Promise(function(adv) {
+                    test.onOutPortData(network.processes.sut.component, 'output', adv);
```
dbooth-boston, line 161: Maybe add a comment like:
// This is in a promise because we need to wait for the first event to finish propagating
// before registering 'fail' on the output port a few lines below.

test/jsupdate-behaviour-mocha.js
```diff
@@ -112,10 +213,115 @@ describe("framework-state", function() {
...
+                    test.sendData(network.processes.once.component, 'input', "once");
+                }).then(function() {
+                    test.onOutPortData(network.processes.sut.component, 'output', done);
+                    test.sendData(network.processes.once.component, 'input', "again");
+                });
+            }).catch(fail);
+        }).should.eventually.have.property('data', "Hello once");
+    });
+    it("should fire updater if upstream updater did nothing, after error, after valid output", function() {
```
dbooth-boston: s/should fire/should not fire/

test/jsupdate-behaviour-mocha.js
```diff
@@ -112,10 +213,115 @@ describe("framework-state", function() {
...
+                return new Promise(function(adv) {
+                    test.onOutPortData(network.processes.sut.component, 'output', adv);
+                    test.sendData(network.processes.upstream.component, 'input', "upstream");
+                }).then(function() {
+                    return new Promise(function(adv) {
+                        test.onOutPortData(network.processes.upstream.component, 'output', adv);
+                        test.sendData(network.processes.upstream.component, 'input', "again");
+                    });
+                }).then(function() {
+                    test.onOutPortData(network.processes.sut.component, 'output', done);
```
dbooth-boston, line 294: Actually the updater should NOT be fired in this case.  Even though the upstream went from valid to error to valid, its state (data) did not change, as signaled by undefined being returned from the updater (case 3).  To be clear, the downstream node should be _notified_ of the status change, but the updater should not fire.  (shouldFireUpdater should be called but return false.)

test/jsupdate-behaviour-mocha.js
```diff
@@ -112,10 +213,115 @@ describe("framework-state", function() {
...
+                    test.sendData(network.processes.upstream.component, 'input', "yet again");
+                });
+            }).catch(fail);
+        }).should.eventually.have.property('data', "Hello upstream");
+    });
+    it("should not fire updater if upstream updater corrected error, but has no state", function() {
```
dbooth-boston, line 300: Good.  The rationale here is that the updater has decided that its own 'undefined' initial state is a good enough 'valid' state for its purposes.

test/jsupdate-behaviour-mocha.js
```diff
@@ -124,6 +330,41 @@ describe("framework-state", function() {
...
+                    fail("Hello " + input);
+                })
+            }).then(function(network){
+                network.graph.addEdge('broken', 'output', 'sut', 'input');
+                test.sendData(network.processes.broken.component, 'input', "brake");
+                setTimeout(done.bind(this, "nothing happened"), 100);
+            });
+        }).should.become("nothing happened");
+    });
+    it("should fire error updater if upstream updater set an error data", function() {
```
dbooth-boston, line 341: good

test/jsupdate-behaviour-mocha.js
```diff
@@ -124,6 +330,41 @@ describe("framework-state", function() {
...
+                sut: jswrapper(function(input) {
+                    done("Hello " + input);
+                })
+            }).then(function(network){
+                network.graph.addEdge('broken', 'error', 'sut', 'input');
+                test.sendData(network.processes.broken.component, 'input', "from broken");
+            }).catch(fail);
+        }).should.become("Hello from broken");
+    });
+    it("should not notify attached error port if an updater explicity sets the same error", function() {
```
dbooth-boston, line 356: good

test/jsupdate-behaviour-mocha.js
jamesrdf: This test is checking that the updater is fired after the error is corrected and the upstream updater returned undefined.

test/jsupdate-behaviour-mocha.js
catch-point, line 294: I added an additional check to ensure the updater did not fire, but the test still is checking that the node produced the earlier result

dbooth-boston:
LGTM

test/jsupdate-behaviour-mocha.js
dbooth-boston: I am probably just not following the code properly, but I do not understand how 'done' would be called at this point if the sut updater does not fire, and at this point the sut updater should not fire.   If the sut updater does not fire, then its output listeners should not be notified -- there should not be an output event for sut's output -- so what would cause 'done' on line 294 to be invoked?

dbooth-boston:
I gave this an LGTM 12 days ago, so I'm not sure what needs to be re-reviewed.  But I did mark a question at jsupdate-behaviour-mocha.js  line 293.   Is more review needed from me on this?  If so, what parts?

glenna-m:
I have fixed the framework/javascript-wrapper/framework issues with 3 of the 4 failing tests from this suite (though have not checked in yet).   

Would like to talk through the final test case: 
_"should produce output, but not fire updater if upstream updater did nothing, after error, after valid output"_
to be sure I understand why we would not want to call the updater in this scenario, but still pass on info.  After clearing the error, I would have expected the updater to be called so it can adjust if need be.

test/jsupdate-behaviour-mocha.js
dbooth-boston, line 294: To clarify the desired behavior, the upsteam node ('upstream') receives input three times, causing its updater to be fired three times.  The first time its updater is fired, the updater returns a value "upstream" that is set into its state.  This state change causes the downstream node ('sut') to be notified and sut's updater is fired because it's input state changed.  The second time 'upstream' receives input upstream's updater is again fired, but this time it throws an error, which does not cause upstream's output.data to change, but it does cause upstream's output.error to be set.  As a result, the downstream node 'sut' should be notified that the status of its input changed, but sut's shouldFireUpdater should return false, because its input is no longer ready.  The third time upstream receives input its updater is again called and this time it returns undefined, signaling that its output.data did not change.  However, because its updater returned without throwing an error, its output.error flag is cleared, causing the downstream node 'sut' to be notified that its input status changed (even though its input data did not change).  This causes sut's shouldFireUpdater to be called, but it should return false because the node's input data did not change, so sut's updater should NOT be called.

This test makes me realize that there are cases when events need to be propagated downstream even if a node's updater does not fire, because the node's state becomes stale or non-stale (i.e., invalid or valid).  Events need to be propagated downstream if either the state's data has changed (as indicated by a new LM) or its status changes (error/non-error or stale/non-stale).  I suppose this means we need another flag -- output.stale -- though maybe there is an easier way to handle it.

catch-point, line 294: The intent of this test is to ensure that the sut notifies downstream nodes that it's state is valid (again), even though its updater was not fired. If that is not appropriate behaviour we should discuss this.

dbooth-boston, line 294: Decision: We will add a 'stale' flag to the state objects, which will be set independently from the 'error' flag.

PR #68, opened 2016-03-28
Always extend state object

catch-point:
Useful when only particular properties of a state need to be changed and not the entire object

### DIFF
--- a/src/input-states.js
+++ b/src/input-states.js
@@ -3,7 +3,7 @@
 var _ = require('underscore');
 
 /**
- * Getter and Setter for InPort states.
+ * Getter and Merger for InPort states.
  *
  * @this a vni.  Not used, but may be returned
  * @param node a node facade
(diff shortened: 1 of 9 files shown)


src/input-states.js
```diff
@@ -208,7 +210,9 @@ function setPortStateBySocketId(node, vnid, portName, socketId, state) {
             node.vnis[vnid].inputStates = node.vnis[vnid].inputStates || {};
             node.vnis[vnid].inputStates[portName] = node.vnis[vnid].inputStates[portName] || [];
         }
-        node.vnis[vnid].inputStates[portName][socketId] = state;
+        // Overwrite existing state object
+        var states = node.vnis[vnid].inputStates[portName];
+        states[socketId] = _.extend(states[socketId] || {}, state);
```
dbooth-boston, line 215: This is merging state properties rather than setting an entirely new state object, whereas the API was previously setting an entirely new state object.  I think will be confusing to have some setters merge and others set the entire object.  If a setter is going to merge then I think it would be clearer to give it a different name in the API, such as mergeState.

glenna-m:
Good change.  Makes the code more readable.  LGTM.

dbooth-boston:
LGTM

src/input-states.js
glenna-m: Yes, this is more clear.

dbooth-boston:
Is more review of this PR needed from me?  I gave it LGTM 12 days ago.

catch-point:
Waiting on PR#65 before merging

PR #72, opened 2016-03-31
Rename vni.node to vni.nodeInstance

catch-point:
Rename vni.node to vni.nodeInstance for consistency with port facade


dbooth-boston:
LGTM

PR #75, opened 2016-04-05
Rdf updaters

catch-point:
Rewrite the RDF components as jsupdaters


components/http-accept.js
```diff
@@ -4,13 +4,12 @@ var _ = require('underscore');
...
-var componentFactory = require('../src/noflo-component-factory');
+var wrapper = require('../src/javascript-wrapper.js');
 
-exports.getComponent = componentFactory({
+module.exports = wrapper({
     description: "Extracts the request body as a String and produces a 202 Accepted response. Optionally, also filters on request content type.",
     icon: 'sign-in',
-    outPorts: _.defaults({
+    outPorts: {
         rejected: {
```
dbooth-boston, line 13: Can the node's errorState be used for this instead of defining an additional noflo output port?

components/http-accept.js
```diff
@@ -19,89 +18,58 @@ exports.getComponent = componentFactory({
...
+        var outPorts = this.nodeInstance.outPorts;
+        if (contentTypeMatches(type, input.req.headers['content-type'])) {
+            if (_.has(input.req, 'body')) {
                 if (outPorts.accepted.listAttached().length) {
-                    pair.res.writeHead(202, "Accepted");
-                    pair.res.write("Accepted\n");
-                    outPorts.accepted.send(pair);
+                    input.res.writeHead(202, "Accepted");
+                    input.res.write("Accepted");
+                    outPorts.accepted.send(input);
```
dbooth-boston, line 46: If possible we should get away from noflo-specific calls.  Is this needed?  Could we use a shim node instead?

components/request-template.js
```diff
@@ -3,124 +3,110 @@
...
-            datatype: 'object',
-            required: true,
-            ondata: function(data) {
-                this.nodeInstance.data = _.defaults(data, this.nodeInstance.data);
-            },
-            ondisconnect: promiseOutput(execute)
+            datatype: 'object'
         }
+    },
+    updater: function(method, url, headers, body, parameters, input) {
```
dbooth-boston: For style I would rather have updaters be  named functions (instead of being defined inline as lambdas), so that they are more recognizable as separable functions, like this:

<pre>
module.exports = wrapper({
  . . . 
  updater: foo
});

function foo(a, b, c) {
  . . . 
}
</pre>

components/http-accept.js
catch-point, line 13: No, this is not used to transmit state, but to provide the HTTP response back over HTTP.

components/http-accept.js
catch-point, line 46: I don't think so as this is dependent on noflo-webserver to work.

components/http-accept.js
dbooth-boston, line 13: ok

dbooth-boston:
LGTM

components/object.js
```diff
@@ -2,48 +2,14 @@
 
 var _ = require('underscore');
 
-var promiseOutput = require('../src/promise-output');
-var componentFactory = require('../src/noflo-component-factory');
+var wrapper = require('../src/javascript-wrapper.js');
```
glenna-m, line 5: I am curious why you felt that we needed to make an object updater.  An updater that wishes to pass an object can already define it and send it with a VNI.  Why was this primarily so your subgraphs could create objects with a VNI associated?  Other?

glenna-m:
Now that I have verified it with the latest pipeline, LGTM.  This should be good to merge.

components/object.js
catch-point, line 5: object.js was created so that fbp files could define objects from property values. I don't quite follow your second sentence. Can you provide more of an example of an alternative?

PR #79, opened 2016-04-16
Adjusted timeout in mocha test of Framework to work on dbooth laptop

dbooth-boston:
Also fixed a minor bug in the test -- I think.  You should look at it and see of you agree.


dbooth-boston:
Oops, wrong branch.  I'll try again.

PR #80, opened 2016-04-16
Adjusted test timeout to work on dbooth laptop

dbooth-boston:
Also fixed a minor bug in the test -- I think. You should look at it and see of you agree.

### DIFF
--- a/test/framework-ondata-mocha.js
+++ b/test/framework-ondata-mocha.js
@@ -430,8 +430,8 @@ describe("framework-ondata", function() {
 
             // Define the fRunUpdater that framework should invoke
             var fRunUpdater = function( vni ) { 
-               // Should not call fRunUpdater when only optional data has been sent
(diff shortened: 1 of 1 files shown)


test/framework-ondata-mocha.js
```diff
@@ -430,8 +430,8 @@ describe("framework-ondata", function() {
 
             // Define the fRunUpdater that framework should invoke
             var fRunUpdater = function( vni ) { 
-               // Should not call fRunUpdater when only optional data has been sent
-               // and required port data is still missing
+               // Should not call fRunUpdater when data is missing 
+               // from an attached port
 console.log('in fRunUpdater');
```
glenna-m: Should probably remove the console line.  We do not need that since the fail will shut it down if the updater executes.

glenna-m:
LGTM

catch-point:
LGTM

PR #81, opened 2016-04-25
Http delivery

catch-point:
Add HTTP delivery component that can serve data received on its input port


components/http-delivery.js
```diff
@@ -0,0 +1,19 @@
...
+var _ = require('underscore');
+
+var wrapper = require('../src/javascript-wrapper.js');
+
+/**
+ * This updater takes response content, response type, and a {req, res} pair as
+ * input. It then writes the content and type to the response and returns the
+ * input pair.
+ */
+module.exports = wrapper(function(content, type, input) {
```
dbooth-boston: Is it a common convention to call the {req, res} pair 'input'?  If not it would seem clearer to me to call it reqRes or something like that.

graphs/http-delivery-server.fbp
```diff
@@ -0,0 +1,24 @@
...
+INPORT=delivery.TYPE:TYPE
+INPORT=webserver.LISTEN:LISTEN
+OUTPORT=delivery.OUTPUT:OUTPUT
+
+# Initial Information Packets (IIP)
+'data' -> KEY data(objects/ExtractProperty)
+'/data/query/vnid' -> GET setVnid
+'/vnid' -> SET setVnid(rdf-components/json-pointer)
+
+# When a request comes in sent it to the delivery node and respond
```
dbooth-boston: s/sent/send/

components/json-pointer.js
```diff
@@ -0,0 +1,50 @@
...
+var _ = require('underscore');
+var jsonpointer = require('jsonpointer');
+
+var promiseOutput = require('../src/promise-output');
+var componentFactory = require('../src/noflo-component-factory');
+
+/**
+ * Takes the value of the get pointer for some input and sets that value on the
+ * input at the set pointer. If there is not set pointer provided, this will
+ * return the get pointer value, otherwise it will return the input object.
```
dbooth-boston, line 12: Philosophical comment: I would rather avoid components that are so fine grained.  This feels like it should be done directly in JS rather than in an explicit node in a pipeline.  I think I see why you are using it, so let's go ahead with it.  But I hope we can figure out a way in the future to avoid cluttering the pipeline with trivial functions like this.  Grouping things into subgraphs would certainly be one way, but there may be others.  The RDF Pipeline Framework was designed for larger scale (coarser) granularity, in contrast with noflo.

test/http-delivery-mocha.js
```diff
@@ -0,0 +1,71 @@
...
+describe('http-delivery', function() {
+    it("should receive a request using a sub-graph", function() {
+        return new Promise(function(done, fail) {
+            test.createNetwork({
+                delivery: "rdf-components/http-delivery-server"
+            }).then(function(network){
+                afterEach(_.once(network.stop.bind(network)));
+                network.graph.addInitial("Hello World!", 'delivery', 'content');
+                network.graph.addInitial('text/plain', 'delivery', 'type');
+                network.graph.addInitial(1337, 'delivery', 'listen');
```
dbooth-boston: I assume 1337 is the port number.  Might be good to define that as a var above, as a means of documenting it.

dbooth-boston:
LGTM subject to minor suggestions

components/http-delivery.js
glenna-m: I think input or in is just common as a generic input parameter that takes data in Noflo.    We can be more explicit in our naming (I am as guilty of this as James in my components).

components/json-pointer.js
glenna-m, line 12: Maybe we should have a subdirectory for some of these helper components like parseJson or json-pointer since they are wrapped like pipeline components but aren't really providing pipeline functionality.

test/http-delivery-mocha.js
glenna-m: Actually, should the port number be configurable?

glenna-m:
LGTM

components/json-pointer.js
catch-point, line 12: json-pointer is written in JS and used by fbp files. fbp files should not be embedding JS directly.

sparql/load-base-url-vnid.ru.hbs
```diff
@@ -0,0 +1,9 @@
+# load-base-url-vnid.ru.hbs
+
+# Loads the graph-url as the data into an optional graph-uri
+# This file uses http://handlebarsjs.com/
+
+LOAD <{{{base-url}}}?vnid={{{vnid}}}&ext=.ttl>
```
dbooth-boston, line 6: good

test/http-delivery-mocha.js
```diff
@@ -0,0 +1,111 @@
...
+                return new Promise(function(cb) {
+                    http.get(origin + '/?vnid=there', cb);
+                });
+            }).then(function(res){
+                res.setEncoding('utf8');
+                res.on('data', done);
+            }).catch(fail);
+        }).should.become("Hello There!");
+    });
+    xit("should round trip jsonld through SPARQL LOAD", function() {
```
dbooth-boston, line 74: good

dbooth-boston:
LGTM

PR #82, opened 2016-04-25
Network stop

catch-point:
Always stop network implicitly


glenna-m:
The changes made look fine - LGTM on that.   I notice there are a few places where the promise is setting up a function(fail) and not actually supplying it, or using it. Do we need this code?  Would Promise(done) be sufficient on these tests?

dbooth-boston:
LGTM

catch-point:
return new Promise(function(done) { ..}) is fine when the second argument (fail) is not used.

PR #83, opened 2016-04-25
Bug fix for undefined to defined data state transition.

glenna-m:
Bug fix for undefined to defined data state transition.  

The only code change in this is line 72 on javascript-wrapper. 

The other changes in this are just white space changes to comply with our agreed upon coding standard.


catch-point:
LGTM

dbooth-boston:
LGTM

PR #84, opened 2016-04-26
Id mapper

glenna-m:
ID mapper component as used by the pipeline

I wrote this component to map regular CHCS patient ids into the AHLTA ids for procedures in our current pipeline data sets. 

The design, however, is general purpose - it could be used to map any attribute into another one.

### DIFF
--- a/components/id-mapper.js
+++ b/components/id-mapper.js
@@ -1,0 +1,43 @@
+ // id-mapper.js
+var _ = require('underscore');
+
+var wrapper = require('../src/javascript-wrapper');
+
+module.exports = wrapper(idMapper);
+
+/**
+ * Maps the specified id to another value using the specified map.
+ * This is useful when mapping from an ID on one system (e.g., CHCS patient ID) to 
+ * an ID from another system, which may or may not even have the same attribute name 
+ * for the id.
+ *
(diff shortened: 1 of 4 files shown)


catch-point:
I see this component as doing two roles: 1) extracts a property, and 2) maps one value into another value. I think an extract property updater would be useful in other situations and should be on its own. The mapping of values could be simplified if it took a hash of keys to values (assuming all IDs are strings). That would eliminate the need for map_keys (a two item array of property names) and make it easier to understand its role.

glenna-m:
Let's talk, James.  I'm not sure splitting it into two components for an extract outweighs the disadvantage of complicating the graph for this use case, but am open to discussion.  It is already working with a hash for mapping the ids, but it sounds like you have other ideas for an attribute hash so I am  interested in hearing more your thoughts on that.  Please let me know when you have a few minutes.

glenna-m:
James & I agreed that extract property is a useful component but it is not related this component since there is no extraction being done here.  I have adopted his flat hashmap suggestion in the latest checkin.

dbooth-boston:
LGTM

catch-point:
LGTM

PR #85, opened 2016-04-26
Xml to rdf lib

glenna-m:
```
New xml-to-ref-lib component & component-helper

This checkin also includes the xlst support files for converting fhir xml to RDF
```


src/javascript-wrapper.js
```diff
@@ -56,19 +56,24 @@ var fRunUpdater = function(updater, updaterArgs, vni) {
...
+         // console.log('updater returned results: ',results);
+
+         if (! _.isUndefined(results)) { 
+             // Got some results back from updater
+             // If the updater returned anything, set the output state with it
+             var newState = vni.outputState();
+             if (newState.lm === oldOutputStateLm || 
+                 _.isUndefined(oldOutputStateLm) && ! _.isUndefined(newState.lm)) { 
+                 // updater did not modify output state lm - create a new output state with new lm
+                 vni.outputState(createState(vni.vnid, results));
```
dbooth-boston, line 74: This should be reusing the state object (instead of creating a new one), and only setting the data and lm fields.  Were you going to fix this in this PR or in a separate PR?

src/javascript-wrapper.js
```diff
@@ -56,19 +56,24 @@ var fRunUpdater = function(updater, updaterArgs, vni) {
...
+
+     }).then(function(results) { 
+         // console.log('updater returned results: ',results);
+
+         if (! _.isUndefined(results)) { 
+             // Got some results back from updater
+             // If the updater returned anything, set the output state with it
+             var newState = vni.outputState();
+             if (newState.lm === oldOutputStateLm || 
+                 _.isUndefined(oldOutputStateLm) && ! _.isUndefined(newState.lm)) {
```
dbooth-boston, line 72: The data field should be set unconditionally here, I believe.  But if the updater also explicitly set the lm already, then I guess we should use that instead of creating a new one:

```
newState.data = results;
if (newState.lm === oldOutputStateLm) { newState.lm = createLm(); }
vni.outputState(newState);  // Not needed, but included for possible future copy-on-write
```

dbooth-boston:
I was thinking further about the way this component does its own event sending, so that it can send out chunks that are small enough for the downstream node to process.    Would it be feasible to instead modify the downstream node to accept an array of chunks, and process one chunk at a time?  No, come to think of it, that would not solve the problem, it would just push the problem downstream.

glenna-m:
It would just push it down stream and it does not scale as well since you have to combine the pieces to send the array.   We already have an array of source XML files coming in here.  I am doing one send per file in this updater.   I could put the translations back into an array (an early implementation did that), but it gets so big that the rdf-load cannot handle it.

src/javascript-wrapper.js
glenna-m, line 74: This was modified in the PR you gave an LGTM to yesterday.  But not the way you are describing here.

The only thing I had modified in that one was handling the undefined corner case. 

Rather than piecemealing these things, why don't you review javascript-wrapper and write up whatever changes you want as issues so we can do a javascript wrapper PR? 

This PR was intended to be for xml-to-rdf-lib component only.

components/xml-to-rdf-lib.js
```diff
@@ -0,0 +1,71 @@
+// xml-to-rdf-lib.js
```
dbooth-boston: good

components/xml-to-rdf-lib.js
```diff
@@ -0,0 +1,71 @@
...
+ * package in the pipeline, node.js will not be interruptable. 
+ *
+ * @this vni context
+ * @param sources array of xml files to be processed; noflo GUI can also send it as a string
+ * @param classpath classpath to transform library
+ * @param transform file path to xslt transform file
+ *
+ * @return FHIR RDF
+ */
+function xmlToRdfLib(sources, classpath, transform) {
```
dbooth-boston: good

lib/xslt/transform.xsl
```diff
@@ -0,0 +1,903 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
dbooth-boston, line 1: good

src/component-helper.js
```diff
@@ -0,0 +1,27 @@
...
+module.exports = {
+
+    /**
+     * Updates the output state LM to a more recent timestamp.
+     * This is useful when a component updater has already handled sending
+     * state and does not want the pipeline framework to do it.
+     *
+     * @param vni
+     */
+    refreshOutputLm: function(vni) {
```
dbooth-boston, line 20: ok

src/noflo-component-access.js
```diff
@@ -63,3 +64,8 @@ function isInPort(port) {
 function isOutPort(port) {
     return port instanceof noflo.OutPort;
 }
+
+function sendIt(data) {
```
dbooth-boston, line 68: ok

test/common-test.js
```diff
@@ -126,6 +126,22 @@ module.exports = {
     },
 
     /**
+     * Get the expected classpath for the saxon jar depending on the current operating system.
+     */
+    saxonClasspath: function() {
```
dbooth-boston, line 131: ok

test/component-helper-mocha.js
```diff
@@ -0,0 +1,56 @@
...
+var helper = require('../src/component-helper');
+
+describe('component-helper', function() {
+
+    it('should exist as an object', function() {
+        helper.should.exist;
+        helper.should.be.an('object');
+    });
+
+    describe('#refreshOutputLm', function() {
```
dbooth-boston, line 24: ok

test/xml-to-rdf-lib-mocha.js
```diff
@@ -0,0 +1,109 @@
...
+
+            expect(compFactory.updater.call(node.vni(''), 
+                                            ['./test/data/testPatient.xml'], 
+                                            classpath,
+                                            './lib/xslt/transform.xsl')).to.not.throw.error;
+        }); 
+   });
+
+   describe('functional behavior', function() {
+       it('should translate xml to RDF in a noflo network', function() {
```
dbooth-boston: ok

dbooth-boston:
LGTM

dbooth-boston:
LGTM

catch-point:
LGTM

PR #86, opened 2016-04-27
Fixing an error state update bug David spotted in javascript-wrapper

glenna-m:


dbooth-boston:
LGTM

src/javascript-wrapper.js
```diff
@@ -82,7 +82,11 @@ var fRunUpdater = function(updater, updaterArgs, vni) {
              outputState.error = true;
              vni.outputState(outputState);
          } 
-         vni.errorState(createState(vni.vnid, e.toString()));
+         
+         var errorState = vni.errorState();
+         errorState.data = e.toString();
+         errorState.lm = createLm();
+         vni.errorState(errorState);
```
catch-point, line 89: This could be simplified to vni.errorState({data: e.toString(), lm: createLm()}); since errorState merges objects.

src/javascript-wrapper.js
```diff
@@ -82,7 +82,11 @@ var fRunUpdater = function(updater, updaterArgs, vni) {
              outputState.error = true;
              vni.outputState(outputState);
          } 
-         vni.errorState(createState(vni.vnid, e.toString()));
+         
+         var errorState = vni.errorState();
+         errorState.data = e.toString();
```
dbooth-boston, line 87: Let's leave this as e instead of turning it into a string.

PR #90, opened 2016-04-30
Montreal Demo

glenna-m:


graphs/cmumps-patient-jsonld.fbp
```diff
@@ -15,7 +15,7 @@ OUTPORT=parseJson.OUTPUT:OUTPUT
 'http://{+authority}/patient_graph{?dataset,datatype,patientid}' -> URL request
 
 # Default request URL parameters
-'chcs-ab' -> VALUE setDataset
+'cmumps-ab' -> VALUE setDataset
```
catch-point, line 18: I get a "No such dataset found: cmumps-ab" when I try and use this.

components/cmumps2fhir-demographics.js
```diff
@@ -0,0 +1,48 @@
+// cmumps2fhir-demographics.js
+
+var _ = require('underscore');
+
+// NOTE: The rdftransforms are under a private git repository belonging to Hokukahu, LTD.  
+// Please contact Hokukahu if you wish to run this component
+var extractor = require('../../rdftransforms/translate/chcs');
```
catch-point: This can cause a lot of headaches setting up. I think we should use an npm module for rdftransforms if they need to be in a separate repository. We cane use optionalDependencies for these sort of modules.

See https://docs.npmjs.com/files/package.json#optionaldependencies

components/fhir-to-xml.js
```diff
@@ -0,0 +1,49 @@
...
+
+   // console.log('\nenter fhir-to-xml with ', fhir);
+   var fhirJsonToXml = new fhir2xml.FHIRConverter();
+   var xmlFileNames;
+
+   if ( _.isArray( fhir )) { 
+           
+      xmlFileNames = _.map( fhir, function( obj ) {
+           // console.log('\nfhir to xml object to convert: ',obj,'\n');
+           var obj2 = _.pick( obj, function(value, key, object) {
```
catch-point, line 27: Both of these condition blocks have common code. Can you put it into a common function to easy maintenance?

components/shex-cmumps-to-rdf.js
```diff
@@ -0,0 +1,123 @@
...
+var util = require('util');
+
+var shexiface = require("../lib/shex/shexiface");
+var wrapper = require('../src/javascript-wrapper');
+
+module.exports = wrapper( updater );
+
+function updater(data) {
+    // shexiface.now("enter updater");
+    parsedData = JSON.parse(data);
```
catch-point, line 23: I think this line would be better off as var parsedData = _.isString(data) ? JSON.parse(data) : data;

components/shex-cmumps-to-rdf.js
```diff
@@ -0,0 +1,123 @@
...
+    
+    parsedData["@context"] = graphContext["@context"];
+
+    // Filter to known types
+    // for cleaned.jsonld, 2.1 w, 4.2 w/o
+    if (true)
```
catch-point: no need for this line

components/shex-cmumps-to-rdf.js
```diff
@@ -0,0 +1,123 @@
...
+var shexiface = require("../lib/shex/shexiface");
+var wrapper = require('../src/javascript-wrapper');
+
+module.exports = wrapper( updater );
+
+function updater(data) {
+    // shexiface.now("enter updater");
+    parsedData = JSON.parse(data);
+    
+    parsedData["@context"] = graphContext["@context"];
```
catch-point, line 25: What if data already has a @context?

catch-point:
This PR references the schema http://hokukahu.com/schema/chcss and uses CHCS in shex filenames. Is that desired for use in this repository?

lib/shex/invoke.js
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env node
```
catch-point, line 1: Why do you want to include this file?

src/component-helper.js
```diff
@@ -0,0 +1,27 @@
...
+     * Updates the output state LM to a more recent timestamp.
+     * This is useful when a component updater has already handled sending
+     * state and does not want the pipeline framework to do it.
+     *
+     * @param vni
+     */
+    refreshOutputLm: function(vni) { 
+        var outputState = vni.outputState();
+        outputState.lm = createLm();
+        vni.outputState(outputState);
```
catch-point, line 23: This could be written in one line as vni.outputState({lm: createLm()});

catch-point:
I think the shex files and xslt files should go into a top level folder and remove the lib folder. I also think the javascript files should not go into a shex folder, but the existing src folder.

lib/shex/CHCS_clinical_chemistry.shex
```diff
@@ -0,0 +1,31 @@
...
+  ^chcs:clinical_chemistry-63 {                                                           # <63_04-6929184.919299_1003681>
+    a [chcs:Lab_Result-63],
+  # chcs:identifier                     LITERAL  %map:{ fm:order.identifier %},          # "63-000007"
+    rdfs:label                          LITERAL,                                         # "DOEXXX,JANE"
+    chcs:patient-63 {
+      chcs:identifier                   LITERAL  %map:{ fm:subject.identifier %} },      # <6-7893413> "DOEXXX,JANE"
+  },
+  a [chcs:Clinical_Chemistry-63_04],
+  chcs:identifier                       LITERAL  %map:{ fm:order.identifier %},          # "63-000007"
+  chcs:login_date_time-63_04            LITERAL  %map:{ fm:apparentDate %},              # "1990-01-01T08:07:01Z"
```
dbooth-boston: eric and glenna, I'd like all timestamps to be 00:00:00 .  I see a couple that are not, in https://github.com/rdf-pipeline/noflo-rdf-components/blob/montreal-demo/lib/shex/CHCS_clinical_chemistry.shex Would one of you please change them and then delete the file history from github?  thanks

lib/shex/CHCS_clinical_chemistry.shex
```diff
@@ -0,0 +1,31 @@
...
+  chcs:specimen-63_04 {
+    rdfs:label                          LITERAL  %map:{ fm:specimen.type %} }*,          # <55-89> "BLOOD"
+  chcs:collection_sample-63_04 {
+    rdfs:label                          LITERAL  %map:{ fm:container.type %} }*,         # <87-432> HIV TUBE
+  chcs:date_time_specimen_taken-63_04   xsd:dateTime %map:{ fm:specimen.receivedTime %}, # "1990-01-01T08:07:01"^^xsd:dateTime
```
dbooth-boston: Another timestamp to zero out

lib/shex/CHCS_result.shex
```diff
@@ -0,0 +1,21 @@
...
+<result> { # <11_07-3618_9999999.836399_000007>
+  a                                     [chcs:Result-63_07],
+  ^chcs:result-63_04 {
+    chcs:identifier                     LITERAL  %map:{ fm:order.identifier %},       # "63-000007"
+  },
+  chcs:test-63_07 {
+    rdfs:label                          LITERAL  %map:{ fm:Observation.code %} },     # <77-2345> "INPROCESSING ABO"
+  chcs:lab_method-63_07 {
+    rdfs:label                          LITERAL  %map:{ fm:Observation.code2 %} },    # portable code <8710-2036> "INPROCESSING ABO ONLY"
+  chcs:enter_date_time-63_07            LITERAL  %map:{ fm:diagRpt.issued %}          # "1990-01-01T10:54:04"^^xsd:dateTime
```
dbooth-boston: Another timestamp to zero out

dbooth-boston:
LGTM

graphs/cmumps-patient-jsonld.fbp
glenna-m, line 18: Yes, Conor is not renaming his datasets and since we needed the montreal demo ready ASAP, I didn't take the time to test removing it entirely since from some discussion, I thought it might cause problems if I did.  However... I had time to test this morning, so I have now removed this default and the component seems to be working as long as you do specify a dataset.  In the next checkin, this line will be removed.

components/cmumps2fhir-demographics.js
glenna-m: Yep.  Made the change for that.

PR #91, opened 2016-05-02
Reuse rdf fbp

catch-point:
Changes the *-jsonld.fbp to call the common variation that doesn't use jsonld to reduce code duplication.


dbooth-boston:
LGTM

glenna-m:
LGTM

PR #94, opened 2016-05-11
cmumps2fhir components

glenna-m:
Cmumps translation components plus the new naming changes for cmumps-patient-jsonld.fbp

### DIFF
--- a/graphs/cmumps-patient-jsonld.fbp
+++ b/graphs/cmumps-patient-jsonld.fbp
@@ -1,4 +1,4 @@
-# chcs-patient-jsonld.fbp
+# cmumps-patient-jsonld.fbp
 
 # Takes a patient ID and returns JSON-LD.
 
@@ -15,7 +15,7 @@ OUTPORT=parseJson.OUTPUT:OUTPUT
 'http://{+authority}/patient_graph{?dataset,datatype,patientid}' -> URL request
 
 # Default request URL parameters
-'chcs-ab' -> VALUE setDataset
+'cmumps-dataset' -> VALUE setDataset
 '10.255.241.50:8080' -> VALUE setAuthority
 'datatype=all' -> IN parameters(adapters/PropStringToObject) OUT -> PARAMETERS request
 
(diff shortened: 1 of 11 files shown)


dbooth-boston:
LGTM.  Thanks for getting this checked in!

graphs/cmumps-patient-jsonld.fbp
```diff
@@ -15,7 +15,7 @@ OUTPORT=parseJson.OUTPUT:OUTPUT
 'http://{+authority}/patient_graph{?dataset,datatype,patientid}' -> URL request
 
 # Default request URL parameters
-'chcs-ab' -> VALUE setDataset
+'cmumps-dataset' -> VALUE setDataset
```
catch-point, line 18: I still get "No such dataset found: cmumps-ab". We either need to stick with chcs-ab as the default or don't include a default at all.

test/cmumps-patient-jsonld-mocha.js
```diff
@@ -20,16 +20,16 @@ describe('chcs-patient-jsonld subgraph', function() {
...
+            network.processes.cmumps.component.outPorts.output.attach(output);
             return new Promise(function(done) {
                 output.on('data', done);
-                network.graph.addInitial('localhost:' + port, 'chcs', 'authority');
-                network.graph.addInitial('1000004', 'chcs', 'patient_id');
+                network.graph.addInitial('localhost:' + port, 'cmumps', 'authority');
+                network.graph.addInitial('1000004', 'cmumps', 'patient_id');
             });
-        }).should.eventually.have.property('data', "/patient_graph?dataset=chcs-ab&datatype=all&patientid=1000004");
+        }).should.eventually.have.property('data', "/patient_graph?dataset=cmumps-dataset&datatype=all&patientid=1000004");
```
catch-point, line 32: This URL doesn't actually work

glenna-m, line 32: We should talk about this one.  Conor is not renaming his end points and we have no end point that is currently OK with the naming requirement for posting here.   I tried removing the set dataset and setting it to '' but neither worked.  What do you suggest?

PR #96, opened 2016-05-17
subgraph to print vni data contents to console

glenna-m:
Useful for demos where you want the summary not the whole VNI dumped.

Package changes look like more than they really are.  Our graphs and component list is getting long.  I sorted the graphs alphabetically to make it easier to scan and just added on this graph.


catch-point:
LGTM

dbooth-boston:
LGTM

PR #97, opened 2016-05-17
RDF Pipeline Splitter

glenna-m:
My first implementation of the RDF Pipeline Splitter

### DIFF
--- a/src/framework-ondata.js
+++ b/src/framework-ondata.js
@@ -63,7 +63,7 @@ module.exports = function(payload, socketIndex) {
 
              new Promise(function( resolve ) { 
                  // Execute fRunUpdater which will also execute the updater
-                 resolve( wrapper.fRunUpdater( vni ) );
+                 resolve( wrapper.fRunUpdater( vni, payload ) );
 
              }).then( function() { 
                  // fRunUpdater/Updater success path
@@ -283,16 +283,11 @@ function handleOutput( port, lastLm, state ) {
  * @param vni 
  * @param errorFlag (optional) if not undefined, the errorFlag value will be set
  */
-function setOutputErrorFlag( vni, errorFlag ) {
+function setOutputErrorFlag(vni, errorFlag) {
 
-    var outputState  = vni.outputState();
-    if ( _.isUndefined( errorFlag ) ) { 
-        outputState.error = ! _.isUndefined( vni.errorState().data );
-    } else { 
-        outputState.error = errorFlag;
-    }
+    var error =  _.isUndefined(errorFlag) ?  ! _.isUndefined(vni.errorState().data) : errorFlag;
 
-    vni.outputState( outputState );
+    vni.outputState({error: error});
 }
 
 /** 
@@ -434,5 +429,5 @@ function dataIsGood(vni) {
  */
 function shouldRunUpdater(vni, isStale) { 
     // TODO: May add different policies in the future.
-    return !isStale && haveAllInputs(vni) && dataIsGood(vni);
+    return (!isStale) && haveAllInputs(vni) && dataIsGood(vni);
 }
--- a/src/framework-ondata.js
+++ b/src/framework-ondata.js
@@ -63,7 +63,7 @@ module.exports = function(payload, socketIndex) {
 
              new Promise(function( resolve ) { 
                  // Execute fRunUpdater which will also execute the updater
-                 resolve( wrapper.fRunUpdater( vni, payload ) );
+                 resolve( wrapper.fRunUpdater( vni ) );
 
              }).then( function() { 
                  // fRunUpdater/Updater success path
--- a/src/javascript-wrapper.js
+++ b/src/javascript-wrapper.js
@@ -19,7 +19,7 @@ var wrapperHelper = require('./wrapper-helper');
  */
 var fRunUpdater = function(updater, updaterFormals, vni) {
 
-    var updaterActuals = wrapperHelper.getUpdaterParameters(vni, updaterFormals);
+    var updaterActuals = wrapperHelper.getUpdaterActuals(vni, updaterFormals);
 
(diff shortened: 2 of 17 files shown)


src/split-wrapper.js
```diff
@@ -0,0 +1,74 @@
...
+var wrapperHelper = require('./wrapper-helper');
+
+/**
+ *  Define a default updater stub function to be used if the caller
+ * did not pass one in.  Simply returns whatever the original input was so it
+ * can be sent through the output port to the next component
+ *
+ * @param vnid_hash hash of vnids and data values 
+ */
+function defaultUpdater(vnid_hash) {
```
catch-point: This is the same as the javascript-wrapper's defaultUpdater. Why override and re-define it?

glenna-m: I put it in as a place holder. Although it looks similar here (one parameter, returned), we are working with a hash here.  For this version, however, we certainly could use the default javascript wrapper updater.  Join is another matter...

src/javascript-wrapper.js
```diff
@@ -64,26 +36,19 @@ var fRunUpdater = function(updater, updaterFormals, vni) {
...
-         if (! _.isUndefined(results)) { 
+         if (! _.isUndefined(results)) {
              // Got some results back from updater
              // If the updater returned anything, set the output state with it
-             var newState = vni.outputState();
-             if (newState.lm === oldOutputStateLm || 
-                 _.isUndefined(oldOutputStateLm) && ! _.isUndefined(newState.lm)) { 
+             var newStateLm = vni.outputState().lm;
+             if (newStateLm === oldOutputStateLm ||
+                 _.isUndefined(oldOutputStateLm) && ! _.isUndefined(newStateLm)) {
```
dbooth-boston, line 44: This comment pertains to old code here, rather than code changed in this PR, but . . . if newStateLm is defined and oldOutputStateLm is UNdefined, then I think it means that the updater explicitly set the lm.  In that case I think it would be better to use that LM instead of generating a new one with createLm().

src/split-wrapper.js
```diff
@@ -0,0 +1,74 @@
...
+             var states = vni.inputStates();
+             var first = Object.keys(states)[0];
+             var groupLm = states[first].lm;  // Get the LM of first port - that will be the group LM
+
+             // Update output state but do NOT set the lm here - we don't want to send this
+             // Sending is done below on each of the vnids and they get their timestamp then
+             vni.outputState({data: results, groupLm: groupLm});  
+            
+             var hash = vni.outputState().data;
+             var lm = createLm(); // Create an lm for the new vnis we will be sending
```
dbooth-boston: Now that I think this through more, I think this lm could be set to the lm that we pulled from the input hash state (which at this point will be the same as the groupLm) instead of generating a new one.

src/wrapper-helper.js
```diff
@@ -0,0 +1,94 @@
...
+
+module.exports = {
+    executeUpdater: executeUpdater,
+    getUpdaterParameters: getUpdaterParameters,
+    groupLm: groupLm,
+    handleUpdaterException: handleUpdaterException,
+    updateOutputState: updateOutputState
+};
+
+function getUpdaterParameters(vni, updaterFormals) {
```
dbooth-boston: Suggest renaming for clarity: getUpdaterParameters --> getUpdaterActuals

src/javascript-wrapper.js
glenna-m, line 44: That check is there to deal with the first time we execute a component's updater.  In that case, the original OutputStateLM will be undefined since there has been no output yet. The newStateLM (which is really an output state lm), is going to have it's first LM value since we just got it above.   In that case, we do want to pass the data downstream, no?

src/split-wrapper.js
glenna-m: OK - I'll make that change now.

src/javascript-wrapper.js
dbooth-boston, line 44: Yes, the data should get passed downstream.  But won't that happen if newStateLm is used as the value?   I would think that framework-ondata would notice that it differs from the old lm (which is undefined), so it would trigger an event to be sent downstream.  It may not matter much, but I was thinking that if the user went to the trouble of explicitly setting the lm in the updater, then maybe the user has a reason.

src/framework-ondata.js
```diff
@@ -63,7 +63,7 @@ module.exports = function(payload, socketIndex) {
 
              new Promise(function( resolve ) { 
                  // Execute fRunUpdater which will also execute the updater
-                 resolve( wrapper.fRunUpdater( vni ) );
+                 resolve( wrapper.fRunUpdater( vni, payload ) );
```
dbooth-boston: Is payload being used by one of the wrapper's fRunUpdaters?  I didn't see it in either the split wrapper or js wrapper.

src/split-wrapper.js
glenna-m: Discussed this with David.  I will modify this code to simply use the javascript-wrapper default updater in the next checkin.

src/javascript-wrapper.js
glenna-m, line 44: It does not on the first request.  I added this check precisely because it doesn't.

src/wrapper-helper.js
glenna-m: Will do

src/framework-ondata.js
glenna-m: It is being used in join. I developed the code at the same time, but split it up for review.   I'll remove it for now for the split review.

src/javascript-wrapper.js
dbooth-boston, line 44: ok, thanks!

dbooth-boston:
LGTM

PR #101, opened 2016-05-19
Adjusting test-timing for recently added test before/after hooks

glenna-m:
The recent before/after hooks and heavier network test use is causing our network-oriented tests to run slower.  Many were pretty frequently exceeding the default mocha 2 second test timeout causing a lot of failures and making it hard to see what was working and what was not. 

This checkin adjusts the most egregious test timeouts so that they are generally passing now on my system at least (your mileage may vary on a slower box @dbooth-boston).


catch-point:
LGTM

dbooth-boston:
LGTM

PR #102, opened 2016-05-20
Demo component fhir-json-to-xml

glenna-m:
Also a couple helpers to work around renaming issues


graphs/cmumps-patient.fbp
```diff
@@ -0,0 +1,31 @@
+# cmumps-patient.fbp
+
+# Takes a patient ID and returns a string.
```
dbooth-boston: Suggest adding a little more comment: What is the purpose of the returned string?  What does it represent?

glenna-m: It depends on what you query with the HTTP request to a REST end-point (usually Conor's for us, but it could be anything from a REST endpoint with an ID).  

This is a modified version of James's original cmumps-patient-jsonld.fbp graph.  His version is calling parse-json directly after getting the REST response.  I don't want to do that given our renaming changes since I have to manipulate the response first before I can call Mike's translation code - it will fail if I do not.  This is one of the issues I want to talk with you about.

glenna-m: Discussed this with David - we agreed that I will remove this component because he is going to have Mike make changes to the translation software issues I was working around with this.

catch-point:
LGTM

dbooth-boston:
LGTM

PR #103, opened 2016-05-24
XML to RDF component & fhir-to-rdf Subgraph

glenna-m:


dbooth-boston:
This looks good.  Just one thing: could you please use the master fakePatient7 from Mike, so that it has the latest modifications?

glenna-m:
The files I am using are based on fakePatient7, but are in the formats I need to unit test components individually.  Mike's patient 7 is in JSON-LD format and include absolutely everything about that fake patient.  I needed xml.  I also needed files that had demographics or prescriptions or procedures extracted for some tests (e.g., the CMUMPS tests).  I can use Mike's files for system tests, but to truly individually unit test pieces, I needed some customization for unit testing components that work on different data formats with different data input.

graphs/fhir-to-rdf.fbp
```diff
@@ -0,0 +1,21 @@
...
+# Subgraph that accepts a fhir javascript object, and invokes the fhir-json-to-xml component 
+# to convert it to xml, and then the xml-to-rdf component to convert it to RDF.
+
+# publish the ports for this subgraph 
+INPORT=fhirJsonToXml.FHIR:FHIR
+INPORT=repeat.IN:OUTDIR
+INPORT=xmlToRdf.CLASSPATH:CLASSPATH
+OUTPORT=xmlToRdf.OUTPUT:OUTPUT
+
+'xslt/transform.xsl' -> TRANSFORM xmlToRdf(rdf-compponents/xml-to-rdf)
```
catch-point: Could a more descriptive file name be used like xslt/fhir-to-rdf.xsl?

glenna-m: OK... what would be more descriptive?  fhir-xml-to-rdf.xsl?  Other thoughts?

dbooth-boston: fhir-xml-to-rdf.xsl sounds descriptive to me

dbooth-boston:
LGTM

catch-point:
LGTM

PR #104, opened 2016-05-24
Pick up new translator & fhir-to-xml repositories

glenna-m:
Also upgrading versions for rdfstore, n3, & jsonld
Adding back some missing graphs and components


mcarifio:
Clever use of require. Now I finally understand why translators' index.js file is the way it is :-)

catch-point:
LGTM

dbooth-boston:
LGTM

PR #105, opened 2016-05-25
Component update metric collection & profile log

glenna-m:
This check in provides a very basic elapsed time profiler for the noflo-rdf-components, to make it possible to figure out where bottlenecks occur.   There are two parts: 
1. A rolling log that follows the component sequence of execution with entries tracking: 
   - latest updater run time
   - average component update time
   - total elapsed time for component
2. Individual component metrics summary - writes a file per component to metrics directory
   summarizing the current statistics.  These files are a lot easier to write analysis scripts against.

### DIFF
--- a/src/pipeline-component-factory.js
+++ b/src/pipeline-component-factory.js
@@ -20,6 +20,7 @@ var nofloCompFactory = require('./noflo-component-factory');
 var promiseOutput = require('../src/promise-output');
 var vniManager = require('./vni-manager');
 var framework_ondata = require('./framework-ondata');
+var profiler = require('./profiler');
 
 /**
  * Creates a noflo-component factory, as described by the nodeDef parameter, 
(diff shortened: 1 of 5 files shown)


src/pipeline-component-factory.js
```diff
@@ -72,7 +73,12 @@ module.exports = function( nodeDef, wrapper, ondata ) {
...
  */
-var pipelineNodeInit = function( wrapper, facade ) { 
-    vniManager( facade );
+var pipelineNodeInit = function(wrapper, facade) { 
+    vniManager(facade);
     facade.wrapper = wrapper || {};
+
+    // Set up the component profiler 
+    profiler(facade);
+    facade.profiler.createLog(__dirname+'/../logs',
```
catch-point, line 82: When packaged as a npm module, I don't think we should expect to be able to write to the node_modules directory.

glenna-m, line 82: Correcting this for the next check in

dbooth-boston:
Missing framework-ondata.js?

Also, do you have a way to globally enable/disable the profiling?  I think we might as well put that in now, while you're in that code.  Not sure what mechanism should be used to pass the option to the server.  Maybe add a noflo-server option?

glenna-m:
Yes, I know about framework-ondata - we discussed this AM.  It will be in the next check in as I promised.

Regarding enabling or disabling profiling, is that the best way to spend time right now?  I thought our focus was to get something done quick to allow diagnosis of the scalable pipeline, not a complete profiling system?

We are tracking only a handful of measurements which are pretty light weight, and the writes are asynchronous for the log (and syncronous for under 10 lines for metrics - I could make that async too) -> there's not a lot of delay or overhead there. 

I would suggest that I add a flag to the profiler to enable/disable it in the code for now in case there is some surprise issue.  However, I do doubt these simple counts and an average is going to be our bottleneck problem in the pipeline.

PR #106, opened 2016-05-27
New simplified profiler design

glenna-m:


src/profiler.js
```diff
@@ -0,0 +1,61 @@
...
+              },
+
+              eventTypes: {
+                  ONDATA_EVENT: 1,
+                  UPDATE_SUCCESS: 2,
+                  UPDATE_ERROR: 3
+              },
+
+              update: function(startTime, eventType) { 
+                   var time = Date.now() - startTime + 1; // ensure we get at least 1 ms
```
dbooth-boston: Is this "+ 1" needed?  Can it be eliminated, for more accuracy?

glenna-m: I added it because without it, it's impossible to reliably test that it gets updated, and because having a bunch of events with no time looks really odd.   I can remove it, but if I do, there is no way to validate that we are processing these things correctly in unit testing since they will generally show up as 0.

dbooth-boston: How about adding a delay to the updater in the unit tests, in order to ensure a non-zero result?

glenna-m: I did this for the profiler tests, but that becomes very ugly in the framework tests.  

If you don't care, I will take it out.  I felt that 1 ms round-up was acceptable for more reliable testing, since there is some time being spent in there, but if you feel it is not, I'll remove it.  I simply will not look at timestamps in the framework testing.

dbooth-boston: Yes, let's take it out.  I would rather have more accurate times.  

Also, maybe this will help in getting more accurate cumulative times: https://developer.mozilla.org/en-US/docs/Web/API/Performance/now

src/profiler.js
```diff
@@ -0,0 +1,61 @@
...
+                  numberOfErrors: 0,
+                  numberOfEvents: 0,
+                  numberOfUpdates: 0,
+                  startTime: Date.now(),
+                  totalErrorTime: 0,
+                  totalProcessingTime: 0,
+                  totalUpdateTime: 0
+              },
+
+              eventTypes: {
```
catch-point: I think the code would be easier to parse if each event type had its own function/entry point. In the same way that loggers have console.error/console.warn it indicate the different log levels, the profiler could have profiler.success(started), profiler.error(started), profiler.notReady(started) to indicate the event type.

src/framework-ondata.js
```diff
@@ -52,14 +55,14 @@ module.exports = function(payload, socketIndex) {
...
-             // TODO: revisit this logic later, per rdf-pipeline/noflo-rdf-pipeline#35
 
+             // TODO: revisit this logic later, per rdf-pipeline/noflo-rdf-pipeline#35
              // save data used to see if we have a different state after running updater
- 
              clearStateData( vni.errorState ); // clear last error state
 
              // Save wrapper to make it accessible from the Promise
              var wrapper = this.nodeInstance.wrapper;
+             var updaterStartTime = Date.now();
```
catch-point: I think it would be good to call a profiler function here, to track the difference between the number of entries vs exits.

glenna-m: Will do

src/profiler.js
glenna-m: I like this idea, but I don't want non-stop profile calls in the framework which is already a little long.  Let met take a look at the code and see what we can do here.

src/framework-ondata.js
```diff
@@ -52,17 +52,19 @@ module.exports = function(payload, socketIndex) {
          if ( ! _.isFunction( this.nodeInstance.wrapper.fRunUpdater ) ) { 
 
              // Don't have a wrapper fRunUpdater 
+             profiler.stopEvent(eventStart);
```
catch-point: I think it would be better to put this into a finally block at the end of the function to ensure it is called even if there is an error in the framework.

src/framework-ondata.js
```diff
@@ -52,17 +52,19 @@ module.exports = function(payload, socketIndex) {
...
              // TODO: revisit this logic later, per rdf-pipeline/noflo-rdf-pipeline#35
              // save data used to see if we have a different state after running updater
              clearStateData( vni.errorState ); // clear last error state
 
              // Save wrapper to make it accessible from the Promise
              var wrapper = this.nodeInstance.wrapper;
-             var updaterStartTime = Date.now();
+
+             profiler.stopEvent(eventStart); // done with general event processing 
+             var updateStart = profiler.startUpdate();
```
jamesrdf: This could be moved into the Promise constructor function (currently line 70 w/"Execute fRunUpdater"). This would ensure the timer does not start until the updater is actually executed. If the system is under heavy load there may be a delay between the event being fired and the updater being executed.

glenna-m: It could be, but I'm not sure that is better.  We want to ensure that all elapsed time is covered - no gaps for any reason ->  Either the promise set up time should count against the event set up or against the updater.    If I move it down inside the Promise (moving that time onto event set up) that feels wrong since the event set up is complete at this point.  Also the code becomes less readable IMHO.  Let's talk about this.

src/framework-ondata.js
glenna-m: I did not do this originally because the intent was to have the event set up time + updater time equal the total time executing the framework-ondata.  However, in thinking this through with your other comment below, it does simplify things if we make the event time the total time spent in this function (stopping it in a finally block) and then moving the updater time to being around the fRunUpdater call.  We lose no information that way since the event set up time is the event time - updater time.   As long as David does not object, I will make this change.

catch-point:
LGTM

src/profiler.js
```diff
@@ -0,0 +1,69 @@
...
+                      this.metrics.totalUpdateTime += stoptime;
+                      this.metrics.averageUpdateTime = 
+                          Math.round(this.metrics.totalUpdateTime/this.metrics.numberOfUpdates);
+                  }
+
+                  this.metrics.totalProcessingTime += stoptime;
+              }
+         } // profiler
+    }); // extend
+} // module export
```
dbooth-boston, line 69: nice

test/framework-ondata-mocha.js
```diff
@@ -653,6 +697,14 @@ describe("framework-ondata", function() {
...
                logBuffer.startsWith('framework-ondata unable to process fRunUpdater results!').should.be.true;
 
+               // Verify metrics are updated on fRunUpdater exception path
+               node.profiler.metrics.should.be.an('object');
+               node.profiler.metrics.averageUpdateTime.should.equal(0);
+               node.profiler.metrics.numberOfErrors.should.equal(1);
+               node.profiler.metrics.numberOfEvents.should.equal(1);
+               node.profiler.metrics.numberOfUpdates.should.equal(0);
+               node.profiler.metrics.totalUpdateTime.should.equal(0);
+
```
dbooth-boston, line 707: good

dbooth-boston:
Excellent!  LGTM

PR #107, opened 2016-05-31
Develop

catch-point:
Push April and May work into the master branch


glenna-m:
LGTM.  

Do you want to pick up your fix for PR 108 as part of this merge too?   Not required, but it might be nice.

PR #108, opened 2016-05-31
Register the empty authorization value with the phaser

catch-point:
The REGISTER/ARRIVE counts were off, this fixes it


glenna-m:
LGTM

PR #109, opened 2016-06-02
Load balancer

catch-point:


test/load-balancer-mocha.js
```diff
@@ -0,0 +1,90 @@
...
+var os = require('os');
+var fs = require('fs');
+var path = require('path');
+var noflo = require('noflo');
+var test = require('./common-test');
+
+describe('load-balancer', function() {
+    var port = 1337;
+    it("should POST input to single target", function() {
+        return new Promise(function(done, fail){
```
glenna-m, line 19: I'm guessing this test is going to need more than the default 2 sec to run when we execute it as part of the mocha suite.  Have you run the full test suite to see if this test completes in time?

test/load-balancer-mocha.js
```diff
@@ -0,0 +1,90 @@
...
+                test.onOutPortData(network.processes.balancer.component, 'output', done);
+                fs.writeFile(targetFileName, "http://localhost:" + port + "/", function(){
+                    process.env['target-list-file'] = targetFileName;
+                    network.graph.addInitial('target-list-file', 'balancer', 'target_list_file_env');
+                    network.graph.addInitial("Hello", 'balancer', 'input');
+                });
+            }).catch(fail);
+        }).should.eventually.have.property('data', "Hello World!");
+    });
+    it("should distribute to multiple target", function() {
```
glenna-m: "target" should be plural

test/load-balancer-mocha.js
```diff
@@ -0,0 +1,90 @@
...
+                fs.writeFile(targetFileName, list.join('\n'), function(){
+                    process.env['target-list-file'] = targetFileName;
+                    network.graph.addInitial('target-list-file', 'balancer', 'target_list_file_env');
+                    network.graph.addInitial({vnid:"A",data:"A"}, 'balancer', 'input');
+                    network.graph.addInitial({vnid:"B",data:"B"}, 'balancer', 'input');
+                    network.graph.addInitial({vnid:"C",data:"C"}, 'balancer', 'input');
+                    network.graph.addInitial({vnid:"D",data:"D"}, 'balancer', 'input');
+                });
+            }).catch(fail);
+        }).should.eventually.have.all.keys("/one", "/three", "/four", "/five");
```
glenna-m, line 90: You have 5 servers, and send 4 pieces of data.  How do you know it will not to two?   Please add a comment explaining this.

test/load-balancer-mocha.js
```diff
@@ -0,0 +1,90 @@
...
+    });
+});
```
glenna-m, line 92: These tests are good as far as they go.  Two more come to mind: 
1. A variation on your first test that floods several messages to a single target to verify that the messages are not dropped and we get response back.
2. Some error testing e.g., what happens if you don't have access the load balance is misconfigured e.g., no auth info, nothing to route to

components/object.js
```diff
@@ -5,11 +5,15 @@ var _ = require('underscore');
 var wrapper = require('../src/javascript-wrapper.js');
 
 /**
- * Creates a object from a key and value
+ * Creates a object from a key and value or array of keys and array of values
  * @param key Property key
  * @param value Property value
```
glenna-m: Should update the javadoc params to reflect they can now be arrays

glenna-m:
LGTM, after the minor comments are addressed.

test/load-balancer-mocha.js
catch-point, line 19: I'm only running it on my machine, so it hard to say how long it will take with different hardware. If we had a build machine things would be different.

glenna-m:
LGTM

components/zip.js
```diff
@@ -0,0 +1,32 @@
...
+        return;
+    return [array[choose(value, array.length)], value];
+});
+
+function choose(value, len) {
+    var code = hashCode(JSON.stringify(value));
+    return (code % len + len) % len;
+}
+
+function hashCode(str){
```
dbooth-boston, line 39: Please add a comment saying where this hashing algorithm came from.

components/zip.js
```diff
@@ -0,0 +1,32 @@
...
+
+var _ = require('underscore');
+
+var wrapper = require('../src/javascript-wrapper.js');
+
+/**
+ * Merges together one of the values of the array with the given value
+ * @param array Array of values to choose from
+ * @param value to include in output
+ * @return an Array consistent of an item from array and value
```
dbooth-boston, line 19: Why not just return an object directly, instead of returning a two-item array that is later turned into an object on line 31 of load-balancer.fbp?

graphs/load-balancer.fbp
```diff
@@ -0,0 +1,37 @@
+# load-balancer.fbp
+
+# Sends the input packet to one of the target URLs listed in the file referenced
+# in the environment variable whos name is sent to TARGET_LIST_FILE_ENV.
```
dbooth-boston, line 4: Why use an intermediate env var?  Why not pass in the filename directly?

components/zip.js
catch-point, line 19: By using an array instead of an object, property names do not need to be provided and thus keeps the code simple and focused on pairing items together.

graphs/load-balancer.fbp
catch-point, line 4: For cross platform portability. This would allow a system administrator to move/copy the deployment without modifying the pipeline parameters. Very useful when deploying between staging and production machines.

components/zip.js
dbooth-boston, line 19: But isn't the first element of the array later being used as the property name in creating the object anyway, using object.js in load-balancer.fbp line 31?

graphs/load-balancer.fbp
dbooth-boston, line 4: Sounds good.

components/zip.js
```diff
@@ -0,0 +1,39 @@
+// zip.js
```
dbooth-boston, line 1: I'm not finding this filename very informative.  How about selector.js?

components/zip.js
```diff
@@ -0,0 +1,39 @@
...
+
+var _ = require('underscore');
+
+var wrapper = require('../src/javascript-wrapper.js');
+
+/**
+ * Merges together one of the values of the array with the given value
```
dbooth-boston: I think this comment would be clearer as something like:
[[
Selects one member of an array based on the hashcode of a given value.  This is used in load balancing, to distribute requests across an array of available servers, where the value is a vnid.   Hashing is used to ensure that the same server is used for subsequent requests for the same vnid (provided that the server list does not change).
]]

components/zip.js
catch-point, line 19: No, object.js takes this two-item array of values and another two-item array of property names and creates an object with two properties.

components/zip.js
catch-point, line 1: I choose this name as the process is a classic convolution function, maps a tuple of sequences into a sequence of tuples, which is more commonly referred to as zip.

dbooth-boston, line 1: ok.  i hadn't run across it.  thanks for enlightening me!

dbooth-boston:
LGTM

PR #112, opened 2016-06-10
Join component & wrapper

glenna-m:


test/join-wrapper-mocha.js
```diff
@@ -0,0 +1,349 @@
...
+
+        describe('functional behavior', function() {
+
+            it("should execute in a noflo network", function() {
+                this.timeout(4000);
+                return test.createNetwork(
+                    { nodeA: 'rdf-components/parse-json',
+                      nodeS: 'rdf-components/simple-splitter',
+                      nodeB: 'rdf-components/to-upper-case',
+                      nodeJ: { getComponent:
```
catch-point, line 144: You can pass the function directly, no need to wrap it in a getComponent property.

src/join-wrapper.js
```diff
@@ -0,0 +1,139 @@
+// join-wrapper.js
+
+// Implements a join wrapper as per RDF-pipeline wrapper definition
+// https://github.com/rdf-pipeline/noflo-rdf-pipeline/wiki/Wrapper-API
```
dbooth-boston, line 4: Good. Let's add a comment: "This is an initial version that we should revisit when we have a clearer use case and design for how split and join should work."

dbooth-boston:
LGTM subject to one comment.  No need for re-review.  thanks!

PR #113, opened 2016-06-14
Translation wrapper

catch-point:
Add translation-wrapper that delegates to javascript-wrapper after splitting input payload

### DIFF
--- a/src/translation-wrapper.js
+++ b/src/translation-wrapper.js
@@ -1,0 +1,38 @@
+// translation-wrapper.js
+
+var _ = require('underscore');
+var wrapper = require('./javascript-wrapper');
+var ondata = require('./framework-ondata');
+var createLm = require('./create-lm');
+var createState = require('./create-state');
+
+module.exports = function(nodeDefOrUpdater) {
+    var nodeDef = _.isFunction(nodeDefOrUpdater) ? {updater: nodeDefOrUpdater} : nodeDefOrUpdater || {};
+    return wrapper(_.defaults({
+        inPorts: _.mapObject(_.extend({input:{}}, nodeDef.inPorts), function(port, portName) {
+            if (portName != 'input') return port;
+            else return _.extend({ondata: onsplit}, port);
+        }),
+    }, nodeDef));
+};
+
+function onsplit(payload, socketIndex) {
+    if (_.isUndefined(payload)) return ondata(payload, socketIndex);
+    var data = _.isUndefined(payload.vnid) ? payload : payload.data;
+    if (!_.isObject(data)) throw Error('Translation wrapper requires a hash on the input port');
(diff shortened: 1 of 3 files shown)


src/translation-wrapper.js
```diff
@@ -0,0 +1,38 @@
+// translation-wrapper.js
+
```
dbooth-boston, line 2: Please add an opening comment saying what this is.

src/translation-wrapper.js
```diff
@@ -0,0 +1,38 @@
...
+            createLm(), // Create an lm for the new vnis we will be sending
+            payload.error,
+            payload.stale,
+            groupLm
+        ), socketIndex);
+    }, this);
+}
+
+function isAssignedToMe(data, vnid) {
+    return !data.translatedBy || data.translatedBy == this.nodeInstance.componentName;
```
glenna-m: Do you really want  == here?

test/translation-wrapper-mocha.js
```diff
@@ -0,0 +1,253 @@
...
+                        done.groupLm.match(/^LM(\d+)\.(\d+)$/).should.have.length(3);
+                        done.lm.match(/^LM(\d+)\.(\d+)$/).should.have.length(3);
+                        done.groupLm.should.not.equal(done.lm);
+                    });
+                });
+            });
+
+       });
+   });
+});
```
glenna-m, line 253: These tests are nearly identical to split-wrapper-mocha.js.  Could you do something like I did for the common-wrapper tests, to share this code?  Maybe a common-split-wrapper?

src/translation-wrapper.js
glenna-m, line 40: Minor, but it would probably be better to call this "translateBy" since it is requesting translation by that component.  It has not been translated yet.

test/common-wrapper.js
```diff
@@ -60,16 +60,16 @@ module.exports = function(wrapper, defNumPorts) {
         it("should accept object with one inPort definition", function() {
             return Promise.resolve({
                 inPorts: { 
-                    myinput: { datatype: 'string', 
+                    input: { datatype: 'string',
```
glenna-m, line 63: Do we really have to hard code the input parameter name?  That feels a bit kludgey

dbooth-boston:
LGTM

glenna-m:
LGTM as soon as the change from translatedBy to translateBy is made.  It would be good to get this merged to develop.

glenna-m:
Do we still need this PR?

catch-point:
This PR is obsolete and should not be merged. Instead an alternative should be developed that uses a in port flag to indicate that the port is expected to receive an array of input that should be treated as if each item was received individually on the port.

PR #115, opened 2016-06-16
Helpful debug tracing

glenna-m:
This check in provides the ability to turn on and off tracing on key modules such as the wrappers, framework, and a number of the components.   Debugging may be turned on with a global flag (the component-helper.js debugAll flag), or on individual files e.g., framework-ondata.js, allowing the developer to control the amount of tracing produced. 

Instrumented components will have at a minimum, output that looks like this: 
Enter component-file.js - functionName() for nodeName(componentName) 

In addition key events such as updater invocation and return, framework success or failure, number of records processed, and files written, are also traced now.

### DIFF
--- a/components/cmumps2fhir-demographics.js
+++ b/components/cmumps2fhir-demographics.js
@@ -1,12 +1,16 @@
 // cmumps2fhir-demographics.js
 
 var _ = require('underscore');
+var util = require('util');
 
 var extractor = require('translators').cmumps;
 var translator = require('translators').demographics;
 
 var cmumps2fhir = require('./cmumps2fhir');
+var compHelper = require('../src/component-helper');
 var wrapper = require('../src/javascript-wrapper');
+
+var debug = compHelper.debugAll || false;
 
 module.exports = wrapper(patientDemographics);
 
@@ -23,6 +27,12 @@ module.exports = wrapper(patientDemographics);
  * @return the patient's demographic data in FHIR format
  */
 function patientDemographics(data, cmumps_file, fhir_file) {   
+
+    if (debug) {
+        console.log( '\nEnter ' + compHelper.formattedNodeName(this.nodeInstance));
+        // console.log('data: ',util.inspect(data,{depth:null})+'\n');
+    }
+
 
     if (_.isUndefined(data)) { 
         throw Error("PatientDemographics requires data to translate!");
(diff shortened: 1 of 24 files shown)


catch-point:
If we need a logging tool, I think we should be using a third-party tool and not role our own.

glenna-m:
This is not a general logging tool.  This is a way of having some basic tracing of the key info we need to debug our environment when working with a complex noflo graph.  I got tired of adding this kind of stuff in every time.  It gives an easy way to get the 4 pieces of info you want (node name, component name, file and function) along with the key log messages I keep having to add back in.  It's just using console.log - with a debug flag to turn it on and off on a global or local.   No matter what tool you use, you have to instrument your code with the key stuff you want to debug.  I didn't try to instrument the world - just the things I have been using a lot in building our end-to-end pipeline.  Take a look - I would like to hear your thoughts.

catch-point:
When moving to production logging everything to the console may not be desirable. I suggest we adopt winston logging instead. It supports log levels and metadata. It also supports filters which would make debugging code easier as only the relevant information would be written.

https://www.npmjs.com/package/winston

glenna-m:
I have read mixed things about Winston, but agree it does address centralized logging in a distributed environment.  I don't have personal experience with it before, but If you recommend it, I'm OK with that.  

However, it is pretty much orthogonal to what is in this PR.  Whether you use Winson logger or console.log, we need the right messages to follow our flow so we can debug.    Replacing console.log(message)  with winston.log('debug', message) is trivial and IMHO should be done in a separate PR when we decide that's at the top of the priority list.  

My focus with this PR was to get checked in the things I have have been using so I can focus on the larger pipeline building tasks right now.   Taking these changes does not in anyway inhibit moving to Winston later, and it provides the messages we will want to use with Winston.

glenna-m:
Comparison of 3 common log packages: 
https://www.loggly.com/blog/three-node-js-libraries-which-make-sophisticated-logging-simplers/

But again, I'd suggest we postpone central logging for a bit.   My goal in this PR is to make it easier to follow our pipeline flow, given a knowledge of our code base - not deploy the ultimate logging solution.   Do you see anything in this PR that would inhibit the ease of going to Winston later, or that would be problematic?

glenna-m:
Seeing  test errors in both framework-ondata and join-wrapper on the new Winston version of this PR.

PR #116, opened 2016-06-16
On IIP change call each relevant vni updater

catch-point:
On IIP change call each relevant vni updater

### DIFF
--- a/src/vni-manager.js
+++ b/src/vni-manager.js
@@ -39,6 +39,20 @@ module.exports = function( node ) {
     _.extend( node, 
         {
 
+            /**
+             * Calls fn for each known vni in the system
+             *
+             * @this node instance
+             * @return node instance node for easy chaining
+             */
+            forEachVni: function(fn, thisArg) {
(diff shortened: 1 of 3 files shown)


catch-point:
Glenna, does change affect join-wrapper?

glenna-m:
join wrapper could have additional inputs with the IIP empty vnid since the parameter list is extensible by the updater writer.

src/vni-manager.js
```diff
@@ -39,6 +39,20 @@ module.exports = function( node ) {
     _.extend( node, 
         {
 
+            /**
+             * Calls fn for each known vni in the system
+             *
+             * @this node instance
+             * @return node instance node for easy chaining
+             */
+            forEachVni: function(fn, thisArg) {
```
dbooth-boston, line 48: For my own edification, why not let the caller (framework-ondata) iterate through the vnids, using _.each, instead of writing a custom forEachVni, which basically does the same thing as _.each?

dbooth-boston:
LGTM

PR #117, opened 2016-06-23
Flag that marks a component as a translator

glenna-m:
Adds a boolean flag that marks a component as a translator or not.

This flag was added to allow the javascript wrapper to add translator specific metadata in the shared fRunUpdater code (both translator-wrapper and javascript-wrapper use the same fRunUpdater code).  

The  metadata is currently just the id property with the fhir id so we can use it later as the graph name during upload.

### DIFF
--- a/src/noflo-component-access.js
+++ b/src/noflo-component-access.js
@@ -21,7 +21,8 @@ function facadeComponent(node) {
         },
         get componentName() {
             return node.componentName;
-        }
+        },
+        translator: node.translator
     };
     return _.extend(facade, {
         inPorts: _.mapObject(_.pick(node.inPorts, isInPort), facadePort.bind(this, facade)),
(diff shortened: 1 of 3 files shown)


dbooth-boston:
Could this be done more like subclassing instead of adding a boolean flag?  I think that would be a little clearer.

glenna-m:
Although you can kind of fake subclasses in Javscript, it's not really a class-oriented language.  That's why you don't see a lot of constructors or any Java-style subclassing in our code.  The translation-wrapper is not really  a subclass of the javascript-wrapper even though it shares some code.  

This flag was added so we could share the fRunUpdater code and get the metadata on it.  Let's look at that together.

dbooth-boston:
Please rename 'translator' flag to isTranslator.  Then LGTM.

PR #118, opened 2016-06-23
Add components to load rdf file and rdf content string

catch-point:


test/http-delivery-mocha.js
```diff
@@ -111,4 +112,49 @@ describe('http-delivery', function() {
...
+        if (!process.env['rdf-auth-file']) {
+            process.env['rdf-auth-file'] = path.join(os.tmpdir(), 'rdf-auth');
+        }
+        var turtle = [
+            '<http://dbpedia.org/resource/John_Lennon> <http://schema.org/spouse> <http://dbpedia.org/resource/Cynthia_Lennon> .',
+            '<http://dbpedia.org/resource/John_Lennon> <http://xmlns.com/foaf/0.1/name> "John Lennon" .'
+        ].join('\n');
+        var file = path.join(os.tmpdir(), 'rdf-john.ttl');
+        fs.writeFileSync(file, turtle, {encoding: 'utf-8'});
+        var john = {
```
dbooth-boston: Where is john used in this test?

test/http-delivery-mocha.js
```diff
@@ -111,4 +112,49 @@ describe('http-delivery', function() {
             });
         }).should.eventually.have.property('data').that.has.property('@id', "http://dbpedia.org/resource/John_Lennon");
     });
+    xit("should round trip rdf file through SPARQL LOAD", function() {
```
dbooth-boston, line 115: Is this actually round tripping Turtle?  I am not seeing where the data is read back from the SPARQL store.  It looks like it is checking for JSON results on line 158, rather than Turtle.

graphs/rdf-sparql-clear-load-file.fbp
```diff
@@ -0,0 +1,21 @@
...
+# Takes RDF filename string and inserts its content into the target-graph (IRI) at the sparql-endpoint.
+# The optional inport auth_file_env is an environment variable name that contains a file name, which contains the base64 encoded username:password to be included for HTTP Basic authentication.
+
+# Exports
+INPORT=sparql.SPARQL_ENDPOINT:SPARQL_ENDPOINT
+INPORT=sparql.TARGET_GRAPH_URI:TARGET_GRAPH_URI
+INPORT=readContent.FILENAME:RDF_FILE
+INPORT=readContent.ENCODING:RDF_ENCODING
+INPORT=sparql.RDF_TYPE:RDF_TYPE
+INPORT=sparql.AUTH_FILE_ENV:AUTH_FILE_ENV
```
dbooth-boston, line 12: Is there already a test case for the authentication?  If not I think it would be good to add one while you're in this code.

catch-point, line 12: The tests market with xit use HTTP basic authentication

test/http-delivery-mocha.js
catch-point, line 115: The SPARQL results are converted into JSON-LD before checking the @id.

dbooth-boston:
LGTM

glenna-m:
Can we choose names that make it easier to find the right sparql load in the UI?   The current naming makes it pretty tough since everything has the same prefix and that's all you see.  You have to click randomly through several and delete the wrong ones, til you get the right one unless you remember the exact spelling.   Something liike File-Sparql-Clear-Load would be more obvious in the GUI

glenna-m:
LGTM on the code since we really should get this merged.

I can live with the current naming, since I know I can just type something like load-file when looking for rdf-sparql-clear-load-file.  But this will be a bit confusing to a novice using the UI since all of those rdf-sparql options look the same. You have to know exactly what you are looking for.

PR #119, opened 2016-06-23
Allow hash to be updated after other inputs

catch-point:


src/join-wrapper.js
```diff
@@ -44,7 +44,7 @@ var fRunUpdater = function(updater, updaterFormals, vni, payload) {
 
     var groupLm = wrapperHelper.groupLm(vni.inputStates());
 
-    if (_.isUndefined(groupLm)) {
+    if (_.isUndefined(payload.groupLm)) {
         // Just received hash (which has no group lm), but no input, so there's nothing to do yet 
         return;
```
glenna-m, line 47: Since you aren't using the local groupLm (from line 45) in this check on line 47 anymore, it should probably be moved to after this check where it would actually be used.

glenna-m:
LGTM, after the change suggested.  I don't need to review it again.

PR #120, opened 2016-06-24
Patient hash component & tests

glenna-m:
Used to set up a translator hash so translation can be split and distributed among multiple translators down stream.

Note that this version is built on top of the latest debug branch and is using Winston logging.  The debug PR should be merged before this PR.


dbooth-boston:
LGTM

catch-point:
LGTM

PR #121, opened 2016-06-24
Hiding expected error message where tests were deliberately executing…

glenna-m:
cleaning up the framework tests so that expected errors under test do not look like errors.


test/framework-ondata-mocha.js
```diff
@@ -634,22 +635,29 @@ describe("framework-ondata", function() {
...
+                setTimeout(function() { 
                                 done() 
                             }, 
                             1000);
 
-            }).then( function( done ) { 
+            }).then(function(done) { 
                // Should go through here after the timeout
+               Error.captureStackTrace.restore();
+            }, function(fail) {
```
catch-point, line 666: Don't swallow the fail message here, re-throw it instead.

catch-point:
LGTM

PR #122, opened 2016-06-24
Flag that marks a component as a translator

glenna-m:
Adds a boolean flag (isTranslator) that marks a component as a translator or not.

This flag was added to allow the javascript wrapper to add translator specific metadata in the shared fRunUpdater code (both translator-wrapper and javascript-wrapper use the same fRunUpdater code).

The metadata is currently just the id property with the fhir id so we can use it later as the graph name during upload.

I have also found this flag useful in debugging since it makes it easy to add a bit of tracing to look only at translators instead of every component in a large pipeline.

### DIFF
--- a/src/noflo-component-access.js
+++ b/src/noflo-component-access.js
@@ -21,7 +21,8 @@ function facadeComponent(node) {
         },
         get componentName() {
             return node.componentName;
-        }
+        },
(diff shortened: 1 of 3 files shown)


dbooth-boston:
LGTM

catch-point:
LGTM

PR #123, opened 2016-06-25
Javascript-wrapper to set id metadata for translators

glenna-m:


src/javascript-wrapper.js
```diff
@@ -44,7 +44,12 @@ var fRunUpdater = function(updater, updaterFormals, vni) {
              if (newStateLm === oldOutputStateLm ||
                  _.isUndefined(oldOutputStateLm) && ! _.isUndefined(newStateLm)) {
                  // updater did not modify output state lm - update it now
-                 vni.outputState({data: results, groupLm: groupLm, lm: createLm()});
+                 if (vni.nodeInstance.isTranslator && ! _.isUndefined(results.id)) {
+                     var id = 'urn:local:'+results.id+":(translator:"+vni.nodeInstance.componentName+")";
```
dbooth-boston: Small request: I think spaces around the '+' operators would make the above line a bit easier for my eyes to parse.

dbooth-boston: For safety I think we might have to URL-encode anything that is going into the urn:local URI, in accordance with https://tools.ietf.org/html/rfc2141 .   We should also check for any other places where user-provided data is put into the URI also.  We should be sure that the URL-encoding function that we use only percent-encodes characters that must be percent-encoded.  Some do more than they should.  For example, according to RFC2141, it looks like '/' will have be percent-encoded, but ':' will not.

We should probably discuss this a little, to sort out where we want to allow arbitrary IDs versus where we want to make them urn:local URNs.

dbooth-boston: One other thing: I think the URN can be like urn:local:myTranslatorNode:42 , i.e., going MSB to LSB from left to right, without the parens and without the word 'translator:'.

dbooth-boston:
LGTM after removing angle brackets and s/graphId/graphUri/.  thanks!

catch-point:
LGTM

PR #126, opened 2016-06-29
Framework changes for propagating metadata down stream

glenna-m:


src/framework-ondata.js
```diff
@@ -456,3 +459,50 @@ function shouldRunUpdater(vni, isStale) {
     // TODO: May add different policies in the future.
     return (!isStale) && haveAllInputs(vni) && dataIsGood(vni);
 }
+
+/**
+ * Append any input state metadata to the outputState so it gets passed on to downstream nodes
```
dbooth-boston: s/Append/Set default output metadata by appending/

Also add comment: This will be called before the updater is called, so that the updater or wrapper can modify this metadata as needed.

src/framework-ondata.js
```diff
@@ -35,18 +35,21 @@ module.exports = function(payload, socketIndex) {
...
+    // Check if it's stale and if so, hande setting output flag and sending it on downstream if appropriate
+    var isStale = handleStaleData(vni, lastInputState, inputState, outputPorts.output);
 
-     // Set the new input state
-     vni.inputStates(portName, socketIndex, inputState);
+    // Set the new input state
+    vni.inputStates(portName, socketIndex, inputState);
+
+    // append any input state metadata to the outputState so it gets passed on to downstream nodes
+    transmitMetadata(vni, inputState);
```
dbooth-boston: Suggestions: 
1. s/transmitMetadata/setDefaultMetadata/ 
2. Would it make sense to move this call inside runUpdater, next to the call of clearStateData, so that it is only called if the updater will actually be run?  I guess it would also have to iterate through all of the input states.   Actually, in the interest of time, let's not make that change right now.   Once we get more experience with passing metadata we might want to make changes in this area later anyway, so we could make that change then.

test/framework-ondata-mocha.js
```diff
@@ -714,5 +714,223 @@ describe("framework-ondata", function() {
...
 
+        it("should propagate input metadata downstream", function() {
+            var testId = "test identifier";
+            var testdata = "more data";
+            return test.createNetwork({
+                node1: {getComponent: factory({inPorts: {input: {}}, 
+                                               outPorts: {output: {}}
+                                             },
+                                             {fRunUpdater: function(vni){ 
+                                                  vni.outputState({id: testId, data: "some data", lm: createLm()});
```
dbooth-boston, line 727: Add comment: Metadata property will be 'id'

dbooth-boston:
Nice.  LGTM subject to minor editorial comments.

test/framework-ondata-mocha.js
glenna-m, line 727: in this case, yes.  The metadata property can have any legal javascript attribute name.  In these tests, I happen to have used the property name id.  Do you prefer something else?

src/framework-ondata.js
```diff
@@ -456,3 +459,51 @@ function shouldRunUpdater(vni, isStale) {
...
+                    // So downstream node can see them all
+                    logger.debug('Framework found multiple metadata values for ',key);
+                    metadata.push(inputState[key]);
+                    metadata = _.uniq(metadata);
+
+                } else if ((!_.isArray(metadata)) && (metadata !== inputState[key])) {
+                    // Metadata is currently a single value and we have two different values for it
+
+                    // Is either the new value or the old value undefined?  If so, keep whatever the new value is
+                    if (_.isUndefined(inputState[key]) || _.isUndefined(metadata)) {
```
catch-point, line 493: This should just check metadata, to avoid the situation when the first input has a value and a second input overrides the output value with it's undefined value.

glenna-m, line 493: I was deliberately allowing someone to reset a metadata value to undefined.  To do that, they would have to use the metadata property name and explicitly set the value for that property to undefined.  It would not happen if they simply don't have the property. 

@dbooth - can you weigh in on what you want functionally here?   Should the user be allowed to write something that deliberately clears an upstream metadata value?

test/framework-ondata-mocha.js
dbooth-boston, line 727: No, that's fine.  I meant: please add a comment to the test saying:
// Set metadata property called 'id'

I am suggesting this only to make it easier to see.  when I first looked, it wasn't clear to me what metadata was being set, because it relies on the user knowing that 'id' is not one of the framework properties.  Actually, now that I think of it, maybe it would a little clearer if it were called 'myId' instead of 'id'.  It would stand out more.

src/framework-ondata.js
```diff
@@ -456,3 +459,51 @@ function shouldRunUpdater(vni, isStale) {
...
+ * Set default output metadata by appending input state metadata to the outputState 
+ * so it gets passed on to downstream nodes.
+ */
+function setDefaultMetadata(vni, inputState) {
+
+    var inputKeys = Object.keys(inputState);
+    var outputState = vni.outputState();
+    var outputKeys = Object.keys(outputState);
+
+    // Walk to list of keys looking for non-standard state keys - those will be metadata.
```
dbooth-boston: s/to/through/

src/framework-ondata.js
dbooth-boston, line 493: Yes, I think the user should be allowed to deliberately clear a metadata value.  If they do not want to change it then they can avoid setting the property at all, rather then explicitly setting it to an undefined value.

dbooth:
@glenna-m please double check your @mentions.

PR #127, opened 2016-06-30
Add Shex wrapper

catch-point:


components/frame-jsonld.js
```diff
@@ -0,0 +1,35 @@
...
+ * frame and context. If a frame is provided it will be used to frame the json
+ * object, parsed using the optional context. If a frame is not provided, but a
+ * context object is provided, it will be used to compact  the json
+ * object, parsed using the context if the json object doesn't have one.
+ * Otherwise, the json object is returned.
+ * @param input RDF JS Interface Graph object
+ * @param frame JSON-LD Frame object
+ * @param context JSON-LD Context object
+ */
+module.exports = wrapper(function execute(json, frame, context) {
```
dbooth-boston: For stylistic reasons, I would rather that the updater ('execute' in this case) be declared outside of the wrapper call.

src/shex-wrapper.js
```diff
@@ -0,0 +1,129 @@
...
+ *    https://github.com/rdf-pipeline/noflo-rdf-pipeline/wiki/Wrapper-API
+ *
+ * @param nodeDef
+ * @param vni a virtual node instance
+ */
+function fRunUpdater(nodeDef, vni) {
+    // Execute the updater on the VNI context
+    return new Promise(function(resolve) {
+        var input = vni.inputStates('input').data;
+        if (nodeDef.preprocess)
```
dbooth-boston, line 61: Interesting.  I guess tacking on a preprocessor seems like a reasonable way to do it.

src/shex-wrapper.js
```diff
@@ -0,0 +1,129 @@
...
+                }
+            }
+    }, {
+        fRunUpdater: _.partial(fRunUpdater, nodeDef)
+    });
+};
+
+/**
+ * RDF Pipeline SHEX wrapper fRunUpdater API as documented here:
+ *    https://github.com/rdf-pipeline/noflo-rdf-pipeline/wiki/Wrapper-API
```
dbooth-boston, line 51: Please add more comment saying what it expects on input, and what it does (with the conversions between JSON and RDF).  The fact that it is doing multiple conversion steps should be called out in the opening comment.

dbooth-boston:
Will this work with chcss: data in addition to cmumps: data?  I assume so, but want to verify.

dbooth-boston:
LGTM

shex/FHIR_DiagnosticOrder.shex
```diff
@@ -92,8 +92,10 @@ PREFIX : <http://hl7.org/fhir/Coding.>
 
 PREFIX : <http://hl7.org/fhir/Patient.>
 <Patient> {
+  a [fhir:Patient];
   :identifier {
     a [fhir:Identifier],
+    fhir:link [<tag:eric@w3.org,2016:PatchMe>],
```
glenna-m, line 98: PatchMe?  What's going on here?

shex/shexiface.js
```diff
@@ -80,33 +45,30 @@ function schemaPromise (key, fromto) {
         });
     }
 
-    return TYPE_TO_SHAPE[key][saveAs];
+    return typeToShape[key][saveAs];
 }
 
-function ShExMapPerson (fromGraph) {
+function ShExMapGraph (fromGraph, typeToShape, base, staticBindings, makeTargetNode, targetFixup) {
```
glenna-m, line 51: I like this signature better.  More clear what's going on.

glenna-m:
Has Eric reviewed this?  Should probably have him do so on this one.

shex/targets/fhir.js
```diff
@@ -0,0 +1,78 @@
+/** patchIdentifierLinks - n3-api-specific execution of this SPARQL:
```
glenna-m, line 1: Can you add a bit of documentation about what's going on in this patching?

src/shex-wrapper.js
```diff
@@ -0,0 +1,131 @@
...
+ *
+ * @param nodeDef
+ *
+ * @return a promise to create the noflo rdf component
+ */
+module.exports = function(nodeDef) {
+    return factory({
+        inPorts: _.isArray(nodeDef.inPorts) ?
+            nodeDef.inPorts.reduce(function(map, obj) {
+                if (_.isObject(obj)) {
```
glenna-m, line 29: Elegant.  Should probably use this elsewhere.

src/shex-wrapper.js
```diff
@@ -0,0 +1,131 @@
...
+            return json;
+    }).then(function(json) {
+        var groupLm = wrapperHelper.groupLm(vni.inputStates());
+        vni.outputState({data: json, groupLm: groupLm, lm: createLm()});
+    }).catch(function(e) {
+        wrapperHelper.handleUpdaterException(vni, e);
+    });
+};
+
+function jsonld_to_n3(json) {
```
glenna-m, line 85: So are we putting everything into a single graph, or generating multiple graphs from ShEx?

test/shex-cmumps-to-rdf-mocha.js
```diff
@@ -0,0 +1,147 @@
...
+        compFactory.should.be.a('function');
+    });
+
+    it('should instantiate a noflo component', function() {
+        var node = test.createComponent(compFactory);
+        node.should.be.an('object');
+        node.should.include.keys('nodeName', 'componentName', 'outPorts', 'inPorts', 'vni', 'vnis');
+    });
+
+    it('should throw an error if no parameters', function() {
```
glenna-m, line 72: These error case tests would be better if they verified we had meaningful error messages, not just some  exception thrown.

src/shex-wrapper.js
dbooth-boston, line 85: Ideally each lab result should  be in a separate graph.  But for the moment they can all go into the same graph, given that they are all produced by the same translator.

src/shex-wrapper.js
```diff
@@ -0,0 +1,131 @@
...
+            }
+    }, {
+        fRunUpdater: _.partial(fRunUpdater, nodeDef)
+    });
+};
+
+/**
+ * RDF Pipeline SHEX wrapper fRunUpdater API as documented here:
+ *    https://github.com/rdf-pipeline/noflo-rdf-pipeline/wiki/Wrapper-API
+ * This component json-ld on the input port and produces json-ld on the output port
```
dbooth-boston: s/component json-ld/component takes json-ld/

dbooth-boston:
LGTM if it works :)

glenna-m:
LGTM

dbooth-boston:
LGTM

PR #128, opened 2016-06-30
Cmumps2fhir component Winston logger changes

glenna-m:
While working on translator wrapper and the new translator component changes, I found that there were a number of changes that were appropriate with the recent move to using Winston.  This PR is to review just those changes, separately from the new development.


dbooth-boston:
LGTM

PR #130, opened 2016-06-30
Add-metadata component and test

glenna-m:


dbooth-boston:
This looks useful, especially for testing.  But I would suggest having a single input port that takes an object, instead of two ports, so that multiple metadata properties can be set at once.

glenna-m:
It can be useful both for test and for building things where you have out of band data you need downstream.  

I thought a bit about your suggestion and agree the object for setting multiple metadata properties at a time will likely be quite useful - it's a good idea.  However, the version I had is also really useful when you know the name of the metadata you want, and you just need to put a value coming dynamically from another node with it.    Both use cases are valid.

I thought about whether I could cleanly do this as one component and came to the conclusion that I can't because if I have two ports, the single object metadata won't make it to the updater, without having to set some kludgy empty string value.  So... 

I am going to provide two components for this: 
- add-metadata which will support the object with one or more properties to be added as you suggest
- add-metadatum which will support the original name/value pair use case that I wrote for, where the value can be dynamically generated from another component while the name may just be an IIP.

dbooth-boston:
LGTM.  Thanks!

PR #131, opened 2016-07-13
A couple simple smoke test graphs to validate a configuration is working

glenna-m:
This is a couple very simple graphs I developed as a quick smoke test to validate a newly installed noflo-server and the noflo-rdf-components are working correctly: 
- **aloha.json** is a simple pure noflo graph with two components:  core/repeat and core/output.  It simply writes the word _aloha_ to the console.   If this doesn't run, nothing will.
- **cmumps2fhir-demographics.json** is  simple pure rdf-components graph with four components: rdf-components/read-content, rdf-components/parse-json, rdf-components/cmumps2fhir-demographics, and rdf-components/vni-data-output. It simply parses a dummy cmumps jsonld file and converts it to fhir. A summary of the results will be written to the console, and the details will be written to /tmp/fhir.txt or /tmp/cmumps.txt.

### DIFF
--- a/graphs/cmumps2fhir_demographics.json
+++ b/graphs/cmumps2fhir_demographics.json
@@ -1,0 +1,107 @@
+{
+  "properties": {},
+  "inports": {},
+  "outports": {},
+  "groups": [],
+  "processes": {
+    "rdf-components/read-content_phpms": {
+      "component": "rdf-components/read-content",
+      "metadata": {
+        "label": "read-content",
+        "x": 432,
+        "y": 180,
+        "width": 72,
+        "height": 72
+      }
+    },
+    "rdf-components/parse-json_mxqdk": {
+      "component": "rdf-components/parse-json",
+      "metadata": {
+        "label": "parse-json",
+        "x": 612,
+        "y": 180,
+        "width": 72,
+        "height": 72
+      }
+    },
+    "rdf-components/cmumps2fhir-demographics_7irxa": {
+      "component": "rdf-components/cmumps2fhir-demographics",
+      "metadata": {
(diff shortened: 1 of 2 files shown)


dbooth-boston:
LGTM.  This looks useful.

graphs/cmumps2fhir_demographics.json
```diff
@@ -0,0 +1,107 @@
...
+    },
+    {
+      "data": "test/data/cmumps-patient7.jsonld",
+      "tgt": {
+        "process": "rdf-components/read-content_phpms",
+        "port": "filename"
+      }
+    },
+    {
+      "data": "/tmp/cmumps.txt",
```
catch-point, line 93: Any chance these external dependencies could be eliminated?

glenna-m, line 93: test/data/cmumps-patient7.jsonld is not external.  The file path here is one of our test data files that our mocha tests depend on.  If it's not there, you don't have a good noflo-rdf-components build and our mocha tests will be failing.

glenna-m, line 93: Regarding the tmp files, it's wherever the user wants to write their cmumps and fhir files to - we are generating those files when the pipeline runs.  The component will run fine without them, but I put it in there so the user can examine the data for themselves.

glenna-m, line 93: This graph was generated from the steps I documented here for teaching purposes: 
https://github.com/hokukahu/documents/wiki/Create-a-Noflo-RDF-Pipeline-Graph

catch-point:
LGTM

PR #132, opened 2016-07-14
Support for throttling large datasets

glenna-m:
This check in provides support for controlling the number of ids fed into a distributed RDF pipeline.   The following graph illustrates a typical configuration that includes the throttle that feeds in the correct number of new packets, a load balancer to distribute the workload, several pipeline instances that process the work, and the throttle-fire graph which listens for pipeline completions, determines the number of packets to be added next, and invokes the throttle. 

throttle-graph

I have taken the approach of giving the user complete control over the number of records they start with using  by using the throttle component's throttle_size argument.  

After a pipeline instance completes processing a record, the last step is to do an HTTP post back to the fire-throttle graph.  When fire-throttle receives a packet, it will invoke the throttle-fire component, which will return a new throttle_size for the number of packets to feed to the load balancer next - usually this number will be much smaller than the initial throttle_size, since we are simply replacing completed ones.  That component could be changed for more interesting algorithms based on load feedback if we want it in the future.


components/throttle.js
```diff
@@ -0,0 +1,93 @@
...
+var logger = require('../src/logger.js');
+var wrapper = require('../src/split-wrapper.js');
+
+/** 
+ * Reads a file with one record per line, and outputs a hash with the 
+ * throttle_size number of entries for processing downstream. For example, 
+ * given a file with values 1, 2, 3, 4 and a throttle size of 2, this 
+ * component would first output a hash that reads {"1": "1", "2": "2"}.  If a 
+ * subsequent throttle size packet of 1 comes in, it would then send 
+ * {"3": "3"}.  One more packet of size 1 would send {"4": "4"}.
```
dbooth-boston, line 17: Is this comment correct?  If throttle size is 2, wouldn't it send both 3 and 4?

components/throttle.js
```diff
@@ -0,0 +1,93 @@
...
+ * @this vni
+ *
+ * @param throttle_size - the number of hash values to be sent 
+ * @param file_envvar an environment variable with the path to the file
+ *                    to be processed.
+ * @param encoding - file encoding; defaults to UTF-8
+ *
+ * @return a hash with the next elements to be processed.
+ */
+module.exports = wrapper(function(throttle_size, file_envvar, encoding) {
```
dbooth-boston: One little style thing: please declare the updater function separately, and give it a name, instead of passing in the lambda to the wrapper directly.   Thanks

dbooth-boston:
LGTM, subject to minor editorial suggestions.  thanks!

components/throttle.js
```diff
@@ -0,0 +1,93 @@
+// throttle.js
+
+var _ = require('underscore');
+var fs = require('fs');
+
+var idIndex = 0;
```
catch-point: What happens when there are multiple throttle nodes in the same network (sharing the same idIndex, but different file_envvar)?

components/throttle.js
catch-point: The behaviour of this components is very similar to noflo-core/SendNext, but SendNext support multiple instances. Could SendNext be used or a design that is similar?

components/throttle-fire.js
```diff
@@ -0,0 +1,38 @@
...
+ * 
+ * @param throttle_size the number of elements the throttle should feed into the
+ *        pipeline now. 
+ * @param input any data input from the upstream nodes that indicates data
+ *        processing is done and the pipeline is ready for a new batch of 
+ *        data.  The content doesn't matter.  Only that it has arrived.
+ *
+ * @return the number of elements the throttle should feed into the pipeline
+ */
+module.exports = wrapper(function(throttle_size, input) {
```
catch-point: This seams to have the same behaviour as noflo-core/Kick - "generates a single packet and sends it to
  the output port" when something it received on its "in" port. Could core/Kick be used instead?

graphs/fire-throttle.fbp
```diff
@@ -0,0 +1,23 @@
...
+webserver(webserver/Server) REQUEST -> INPUT accept(rdf-components/http-accept)
+accept ACCEPTED -> IN response(webserver/SendResponse)
+accept REJECTED -> IN response
+
+# Send whatever was received from the webserver as input to throttle-fire
+accept(rdf-components/http-accept) OUTPUT -> INPUT throttle(rdf-components/throttle-fire)
+'1' -> THROTTLE_SIZE throttle(rdf-components/throttle-fire)
+
+# Error handling echos errors to console.log
+accept ERROR -> IN error(core/Output)
```
catch-point, line 21: What about errors from throttle?

components/throttle.js
glenna-m, line 17: Comment is correct. 

If throttle-size is two, it will send the next two elements. Since we are just starting, it will send {"1":"1", "2":"2"}.  If it stays two, when it fires again, it would send {"3":'3", "4":"4"}.  The next elements to send are wherever it left off.

components/throttle.js
glenna-m: Expectation is that the throttle and throttle-fire are singletons that control the feeding of data in the network.  I wrote this because I don't want to flood the load balancer with a couple hundred thousand patient records to process.  This lets us take a batch, process, and add more when we are ready for it in a distributed environment - in this case, we are working with 10 VMs.

components/throttle.js
dbooth-boston, line 17: Ah yes.  Sorry, I mis-read it the first time.

components/throttle-fire.js
glenna-m: I thought about that, but didn't think so.    Ideally, I want to fill the pipeline with a lot of patient records to start with - at least one or two per VM.   That means we need to start with a throttle size of 10 or 20.  However, I don't want to continue to send 10 or 20.  I want to simply replace whatever just finished processing with another ID or two, and I wanted to make the next throttle size controllable so we could have different algorithms for what is sent next in the future. Kick just does a blind fire.  It doesn't let me send a value, AFAIK.

graphs/fire-throttle.fbp
glenna-m, line 21: Not a whole lot that can actually go wrong, but I'll put something in for that.  It's good to cover it.  Thanks.

components/throttle.js
glenna-m: SendNext (and Kick) are about firing off and doing the exact same thing.  There is no setting on them to let you send a parameter to the next component.  My use case is not to just kick it or send a single packet like send next does.    The use case here is to start with an initial large throttle_size and then switch to a smaller size as we finish up.  This approach gives us flexibility for dynamically controlling how much we feed into the pipeline instead of just blindly always sending one.

components/throttle.js
catch-point: I think it would be better if variables were local to the component instance and not global.

components/throttle-fire.js
catch-point: AFAIK, the data port on core/Kick is used to populate the packet sent on "out" port, while the "in" port is used to trigger it.

components/throttle.js
glenna-m: Agree.  Will make this change for next checkin.

components/throttle-fire.js
glenna-m: Confirmed this.  Will make this change for next check in.  Good suggestion.

glenna-m:
Merging this since I have addressed all code review comments and we need easy distribution to the LPI Pipeline.

dbooth-boston:
LGTM

PR #133, opened 2016-07-16
Reapplying profiler changes.

glenna-m:


glenna-m:
Merging this down with verbal permission from David & James

PR #134, opened 2016-07-20
Ensuring translator flag and graphUri is set on the translator compon…

glenna-m:
Catching the translators up to the changes made in javascript-wrapper and framework-ondata awhile back.


dbooth-boston:
LGTM

catch-point:
LGTM

PR #135, opened 2016-07-23
Fhir queries

dbooth-boston:
These queries are used and saved in AllegroGraph.  If you update any in AllegroGraph, please check in the changes to github also.


catch-point:
LGTM

glenna-m:
David, these are all using the older urn syntax of urn:test, but more recently you had me switch to using urn:local - do these need updating to the newer urn syntax?

dbooth-boston:
Fixed!  Thanks Glenna.

PR #136, opened 2016-07-26
cmumps/fhir id format change + graphURI metadata isolated to cmumps/f…

glenna-m:
This PR has two changes that are related and requested at the same time: 
1. The Cmumps to FHIR translators were changed to use a different id format. This affected the naming of the graphURI and output files that were using it.   Modifications were made to ensure the naming would keep the previous format of urn:local:fhir:resource-type:id where appropriate. 
2. In addition, David asked that the graphUri metadata be isolated to the cmumps2fhir translation path only.  Previously it had been in the javascript-wrapper.  Moving it out of there involved a fair bit of modification to tests as well as the way it was invoked.


catch-point:
LGTM

dbooth-boston:
LGTM

PR #137, opened 2016-07-26
Ag sparql federate

catch-point:


dbooth-boston:
LGTM

test/ag-sparql-federate-mocha.js
```diff
@@ -0,0 +1,66 @@
...
+    var origin = "http://localhost:" + port;
+    it("should create and close a session with authentication", function() {
+        this.timeout(5000);
+        var requests = 0;
+        var session;
+        return new Promise(function(done, fail) {
+            var server = http.createServer();
+            afterEach(_.once(server.close.bind(server)));
+            server.on('request', function(req, res) {
+                requests++;
```
glenna-m, line 27: I thought the point of federating was to work across multiple servers within a cluster?  This is single server though?  This looks to me like it tests file access using the API, but does not actually test the distributed aspect.  Am I reading this right?  Should we have a test that verifies we actually get the data from a couple servers or is that not possible in a mocha test?   Have we tried this kind of thing on our 10 node cluster?

graphs/ag-sparql-federate.fbp
```diff
@@ -0,0 +1,51 @@
+# ag-sparql-federate.fbp
+
+# Executes a SPARQL query against a dynamic federation of AllegroGraph repositories
+
```
glenna-m, line 10: Can you add a summary of the steps required to do a federated query or point to the documentation of an example?   Is there a manual way to execute a federated query?   Would be good to know how to verify a query while building a graph with  this component.

glenna-m:
LGTM

PR #138, opened 2016-07-27
Confluence component

glenna-m:
This component takes any number of input edges, and waits until all have been received before proceeding.  The inputs will be returned as an array of whatever values were received.   It is called confluence because it behaves like a confluence of rivers. 

This component is useful when you need to synchronize multiple inputs before proceeding downstream with processing.

### DIFF
--- a/components/confluence.js
+++ b/components/confluence.js
@@ -1,0 +1,33 @@
+// confluence.js
+
+var _ = require('underscore');
+
+var wrapper = require('../src/javascript-wrapper');
+
+module.exports = wrapper({ 
+                           description: "Waits for input from all edges and returns an array of the values",
+                           icon: 'cogs',
+                           inPorts: {
+                               input: {
+                                   datatype: 'all',
+                                   multi: true
+                               }
+                           },
+                           updater: confluence
+                        });
+
+/**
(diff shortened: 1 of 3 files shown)


dbooth-boston:
LGTM

components/confluence.js
```diff
@@ -0,0 +1,33 @@
...
+ * acting much like the confluence of rivers.
+ * 
+ * @this vni context
+ * @param input the input port to which all edges should be attached.
+ * 
+ * @return an array with all of the input values
+ */
+function confluence(input) {
+
+    var result = _.isArray(input) ? input : [input];
```
catch-point, line 31: Is what case would this not be an array?

glenna-m, line 31: If there is just an IIP.   I tested that and did not see an array on that path.

catch-point, line 31: I think that behaviour is a bug, no?

glenna-m, line 31: Perhaps.   My focus was on making the pipeline work right now, so I did not pursue it further.  The defensive code ensured this component should work regardless.

catch-point, line 31: At least a FIXME/TODO is needed to document the potential bug

PR #139, opened 2016-07-27
Add a slight delay to ensure a non-zero average update time

glenna-m:
Minor change to add a 1 ms delay to ensure average updater time should be greater than 1 ms in one of the tests so we can verify the value is updated.

### DIFF
--- a/test/profiler-mocha.js
+++ b/test/profiler-mocha.js
@@ -172,6 +172,11 @@ describe("profiler", function() {
                nodeInstance.profiler.stopEvent.should.be.a('function');
                nodeInstance.profiler.startUpdate.should.be.a('function');
                nodeInstance.profiler.stopUpdate.should.be.a('function');
+               function delay(ms) {
+                  ms += new Date().getTime();
+                  while (new Date() < ms){}
+               }
+               delay(1); // wait long enough to ensure average update time is greater than 0
                return 'success';
            };
 


dbooth-boston:
LGTM

test/profiler-mocha.js
```diff
@@ -172,6 +172,11 @@ describe("profiler", function() {
                nodeInstance.profiler.stopEvent.should.be.a('function');
                nodeInstance.profiler.startUpdate.should.be.a('function');
                nodeInstance.profiler.stopUpdate.should.be.a('function');
+               function delay(ms) {
+                  ms += new Date().getTime();
+                  while (new Date() < ms){}
+               }
+               delay(1); // wait long enough to ensure average update time is greater than 0
```
catch-point, line 179: Can this be changed to use common-stubs.js#promiseLater?
https://github.com/rdf-pipeline/noflo-rdf-components/blob/master/test/common-stubs.js#L31

glenna-m, line 179: I like that is less active, but not sure it's worth the effort for a test with only a 1 ms delay right now.

test/profiler-mocha.js
```diff
@@ -172,6 +172,11 @@ describe("profiler", function() {
                nodeInstance.profiler.stopEvent.should.be.a('function');
                nodeInstance.profiler.startUpdate.should.be.a('function');
                nodeInstance.profiler.stopUpdate.should.be.a('function');
+               function delay(ms) {
+                  ms += new Date().getTime();
+                  while (new Date() < ms){}
```
catch-point, line 177: This will always return false even if the clock has not advanced. This may not be noticeable on unix, but I know on Windows the JVM's millisecond clock (as a comparison) is not as accurate as on unix.

glenna-m, line 177: Interesting.  On the Mac this works with no problem.

catch-point:
LGTM

glenna-m:
Closing this since I implemented a better solution as part of PR 142

PR #140, opened 2016-07-27
Add a default throttle-size of 1 to the throttle component.

glenna-m:


dbooth-boston:
LGTM

components/throttle.js
```diff
@@ -39,15 +39,13 @@ module.exports = wrapper(throttler);
  * @return a hash with the next elements to be processed.
  */
 function throttler(throttle_size, file_envvar, encoding) {
-
     var node = this.nodeInstance;
     if (_.isUndefined(node.throttle)) node['throttle'] = {}; 
 
     // Get the number of hash elements being requested
     var throttleSize = parseInt(throttle_size);
```
catch-point: This could also be written as var throttleSize = _.isFinite(throttle_size) ? +throttle_size : 1;

catch-point:
LGTM

components/throttle.js
glenna-m: That does look a little neater.  I can make that change and commit.

PR #141, opened 2016-08-03
Throttle requests

catch-point:
Only allow one request-template to be active at once


dbooth-boston:
LGTM

glenna-m:
LGTM based on reading the code.  Will try it out today.

glenna-m:
This looks to have fixed the max open file limit issue.  It would be good to get it merged.

PR #142, opened 2016-08-03
Adding componentName to states for easier support/debugging

glenna-m:


src/create-state.js
```diff
@@ -22,7 +22,7 @@ var createLm = require('./create-lm');
  *
```
dbooth-boston, line 23: @param componentName the name of the component to which this state belongs

dbooth-boston:
LGTM after adding one comment.  No need for re-review.

dbooth-boston:
James is off today, and we are running out of time, so please go ahead and merge this.

PR #143, opened 2016-08-03
New Funnel component

glenna-m:


components/funnel.js
```diff
@@ -0,0 +1,86 @@
...
+
+var stateFactory = require('../src/create-state.js');
+var logger = require('../src/logger.js');
+var wrapper = require('../src/javascript-wrapper.js');
+
+module.exports = wrapper(funnel);
+
+/**
+ * Funnels a single input into the pipeline one at a time, and queues the rest
+ * Fires the next one when it receives an empty file name
```
dbooth-boston: @param input Either a patientId, to be queued and sent to the output when it is its turn, or a _____[fill in something appropriate here]____ from a feedback connection, indicating that the previously running patientId finished and the next one can be sent to the output.  The input port should NOT be 'multi' but should be connected _both_ to an upstream node that supplies the patientIds to be processed, and to a feedback node at the end of the pipeline that indicates when the pipeline has finished processing a patientId.   NOTE: This will need to be changed if non-'multi' nodes become prohibited from having multiple connections.

dbooth-boston:
LGTM after added a little more comment (noted).  No need for further review, just go ahead and merge please.

components/funnel.js
```diff
@@ -0,0 +1,103 @@
...
+        node.funnel.executing = '';
+    }
+
+    var funnel = node.funnel;
+    var funnelInput = input;
+
+    // Input coming over http will have extraneous leading and trailing quotes embedded - remove 'em
+    if (!_.isEmpty(funnelInput)) {
+        if (_.isString(funnelInput) && funnelInput.charAt(0) === '"' && funnelInput.charAt(input.length -1) === '"') 
+            funnelInput = funnelInput.substr(1, funnelInput.length -2);
```
catch-point, line 59: What about backslash escaping in the quoted string?

components/funnel.js
```diff
@@ -0,0 +1,103 @@
...
+    }
+
+    if (funnel.executing === funnelInput) {
```
catch-point, line 62: What would happen if the exact same object is indented to be queued twice, like when a patient record changes again before an earlier change has finished being processed?

catch-point:
Why didn't you use core/SendNext for this? It seems very similar.

https://github.com/noflo/noflo-core/blob/0.3.1/components/SendNext.coffee

components/funnel.js
glenna-m, line 59: I am only removing the first and last character if they are quotes.  I do this because when I get the patient ID from the http accept server, it is quoted, even though the original had no quotes.  I'm not replacing embedded quotes which might be escaped.  There cannot be any backslashes on a first quote character since it is the first character.  And if there is on the last one, something is wrong - it doesn't match the first quote.

components/funnel.js
glenna-m, line 62: It would not be queued and would not be processed in that case.   Fair point.  I should probably allow it to queue in that case.

glenna-m:
I tried to use SendNext before writing this.  I spent over an hour trying to make it process and fire reliably and it never did.  It needed both kick and load balancer input and making that work with the feedback loop just was not reliable for me.   The two inputs meant things were always hanging and it didn't know when things were complete.  This funnel is putting metadata on the VNI that tells me when a patient id is completed and it works consistently with that.

It is not enough for the funnel to know it has input. It has to be able to distinguish between new patient ids to be processed, versus notification that a specific patient id has finished processing.

components/funnel.js
catch-point, line 59: Why not use JSON.parse? That would convert "I think there is a \"bug\". Do you?" into the original string (decoding the backslashes).

catch-point, line 59: Would it help if I modified http-accept to run JSON.parse?

glenna-m, line 59: It does not see this as JSON because it is coming in as a single field e.g., "2-000007" .  The JSON parser expects a key value pair.  I tried feeding it values I am getting from http-accept -server, like those above, and the JSON parser threw an exception. 

If you have a key value pair in http-accept-server and are stripping off just the value, then yes, that would be helpful.  The stripping first and last quote if they are present is a hack, but it was a necessary one with the data I was getting and it's pretty safe - I am stripping off the quotes only if there is both a leading and trailing quote.

catch-point, line 59: I have created PR#146 that can be used instead of http-accept-server and will remove these quotes in the output

components/funnel.js
```diff
@@ -0,0 +1,108 @@
...
+
+var stateFactory = require('../src/create-state.js');
+var logger = require('../src/logger.js');
+var wrapper = require('../src/javascript-wrapper.js');
+
+module.exports = wrapper(funnel);
+
+/**
+ * Funnels a single input into the pipeline one at a time, and queues the rest
+ * Fires the next one when it receives the id that has just finished executing.
```
dbooth-boston, line 14: Add comment: WARNING: This will only work when the input is guaranteed to be unique, i.e., never receiving the same ID more than once.  When possible, it should be modified to better distinguish between the two port connections, perhaps by sniffing the data to distinguish them.

PR #144, opened 2016-08-04
Translation Wrapper changes for carrying metadata downstream

glenna-m:


src/translation-wrapper.js
```diff
@@ -32,14 +32,14 @@ function onsplit(payload, socketIndex) {
...
-            vnid,
-            element.data,
-            createLm(), // Create an lm for the new vnis we will be sending
-            payload.error,
-            payload.stale,
-            groupLm
-        ), socketIndex);
+        var state =  createState( vnid,
+                                  element.data,
+                                  createLm(), // Create an lm for the new vnis we will be sending
```
dbooth-boston: createLm() should be called only once, outside the _.each loop, so that the same LM is assigned to each new state.  See split-wrapper.js line 53: https://github.com/rdf-pipeline/noflo-rdf-components/blob/develop/src/split-wrapper.js#L53

dbooth-boston:
One small code change, then LGTM.  No need for re-review.

catch-point:
LGTM

PR #145, opened 2016-08-04
Merge Jun 10 develop code into Master

catch-point:


dbooth-boston:
LGTM

glenna-m:
Assuming you re-ran all unit tests and they pass.  If so, LGTM

catch-point:
There are some timeout issues at the end when run all together, but those pass when run by themselves.

PR #146, opened 2016-08-05
Http accept json

catch-point:
This only accept JSON and produces the JSON object in the output, but otherwise is a lot like http-accept-server.


dbooth-boston:
LGTM

glenna-m:
LGTM

PR #147, opened 2016-08-05
Removing require component-helper from xml-to-rdf

glenna-m:
Current develop branch has an error because xml-to-ref was referencing component-helper
but the recent request throttle merge deleted that file.  Cleaning this up so everything runs cleanly.


dbooth-boston:
LGTM

PR #149, opened 2016-08-09
Support for transient data nodes

glenna-m:
Support for transient nodes

### DIFF
--- a/src/framework-ondata.js
+++ b/src/framework-ondata.js
@@ -312,13 +312,18 @@ function handleOutput( port, lastLm, state ) {
 
         // Got any edges out of this port? 
         if (port.listAttached().length > 0) {
+            var nodeInstance = port.nodeInstance;
             logger.debug('sending state', {
                 state: util.inspect(state,{depth:1}),
-                nodeInstance: port.nodeInstance
+                nodeInstance: nodeInstance
             });
 
             port.send( state );
             port.disconnect();
+
+            if (nodeInstance.isTransient) { 
+                 nodeInstance.deleteAllVnis();
+            }
             return true;
         }  
     } 
--- a/src/framework-ondata.js
+++ b/src/framework-ondata.js
@@ -115,7 +115,7 @@ function runUpdater(nodeInstance, vni, isStale, payload, profiler) {
                     lastOutputLm = undefined;  // got a new error so force sending output
                  }
 
-                 handleOutput(outputPorts.output, lastOutputLm, vni.outputState());
+                 handleOutput(vni, outputPorts.output, lastOutputLm, vni.outputState());
                  handleError(vni, outputPorts.error,  lastErrorState);
 
                  profiler.stopUpdate(updateStart, vni.outputState().error);
@@ -133,7 +133,7 @@ function runUpdater(nodeInstance, vni, isStale, payload, profiler) {
                  stateChange(vni.outputState, lastOutputLm, false);
                  stateChange(vni.errorState, lastErrorState, true);
 
-                 handleOutput(outputPorts.output, lastOutputLm, vni.outputState());
+                 handleOutput(vni, outputPorts.output, lastOutputLm, vni.outputState());
                  handleError(vni, outputPorts.error, lastErrorState);
                   
                  // If we haven't already processed the rejected error, do it now
@@ -213,7 +213,7 @@ function handleError( vni, port, lastErrorState  ) {
            vni.errorState( state );
          } 
 
-        if ( (! handleOutput( port, lastErrorState.lm, state )) && state.data )  { 
+        if ( (! handleOutput(vni, port, lastErrorState.lm, state)) && state.data)  { 
             // State was not sent on to an attached port, and we do have error data
             // so go ahead and log it for debugging/support use.
             if ( state.data.stack ) { 
@@ -270,7 +270,7 @@ function handleStaleOutput(vni, outputPort, lastInputStateLm) {
     var outputState = vni.outputState();
     outputState.stale = true;
     vni.outputState(outputState);
-    handleOutput(outputPort, lastInputStateLm, vni.outputState());
+    handleOutput(vni, outputPort, lastInputStateLm, vni.outputState());
 }
 
 /**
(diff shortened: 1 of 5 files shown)


src/framework-ondata.js
```diff
@@ -312,13 +312,18 @@ function handleOutput( port, lastLm, state ) {
...
                 state: util.inspect(state,{depth:1}),
-                nodeInstance: port.nodeInstance
+                nodeInstance: nodeInstance
             });
 
             port.send( state );
             port.disconnect();
+
+            if (nodeInstance.isTransient) { 
+                 nodeInstance.deleteAllVnis();
```
dbooth-boston: This should only delete the current VNI.  I think I initially said it should delete all of them, but I realized later that it should only delete the current one, because the others may have received input on some ports but not all, and thus may not have fired yet.  Sorry about that!

test/framework-ondata-mocha.js
```diff
@@ -950,5 +950,119 @@ should.be.rejectedWith('No wrapper fRunUpdater function found!  Cannot run updat
...
+
+                     // Verify VNIs have been cleared
+                     expect(nodeInstance).to.not.be.empty;
+                     nodeInstance.vnis.should.be.an('object');
+                     nodeInstance.vnis.should.not.be.empty;
+                 });
+            });
+        });
+
+        it("should clear transient node VNIs after sending output state downstream", function() {
```
dbooth-boston: s/VNIs/VNI/

test/framework-ondata-mocha.js
```diff
@@ -950,5 +950,119 @@ should.be.rejectedWith('No wrapper fRunUpdater function found!  Cannot run updat
...
+                 return new Promise(function(done, fail) {
+
+		     test.onOutPortData(testNode, 'output', done);
+		     test.onOutPortData(testNode, 'error', fail);
+
+                     network.graph.addInitial("init data", 'testNode', 'input');
+                 }).then(function(done) {
+                     test.verifyState(done, '', testData);
+
+                     // Verify VNIs have been cleared
```
dbooth-boston: s/VNIs have/VNI has/

test/framework-ondata-mocha.js
```diff
@@ -950,5 +950,119 @@ should.be.rejectedWith('No wrapper fRunUpdater function found!  Cannot run updat
...
+                     expect(nodeInstance).to.not.be.empty;
+                     nodeInstance.vnis.should.be.an('object');
```
dbooth-boston, line 1019: This will need to be changed to: (a) use multiple VNIs; and (b) verify that the current VNI was cleared, but other VNIs were not changed.

test/framework-ondata-mocha.js
```diff
@@ -950,5 +950,119 @@ should.be.rejectedWith('No wrapper fRunUpdater function found!  Cannot run updat
...
+                     nodeInstance.vnis.should.be.empty;
+                 });
+            });
+        });
+
+        it("should clear transient node VNIs after sending the output state without affecting downstream node state", function() {
```
dbooth-boston: This test also needs to be updated to use multiple VNIs and verify that only the current one is cleared when output is sent.

test/noflo-component-factory-mocha.js
```diff
@@ -226,4 +226,33 @@ describe('noflo-component-factory', function() {
             test.sendData(node, 'input', "isTranslator?");
         }).should.become("isTranslator? true");
     });
+    it("should have nodeInstance isTransient flag that defaults to false", function() {
```
dbooth-boston, line 229: good

test/noflo-component-factory-mocha.js
```diff
@@ -226,4 +226,33 @@ describe('noflo-component-factory', function() {
...
+                        ondata: function(payload) {
+                            done(payload + ' ' + this.nodeInstance.isTransient);
+                        }
+                    }
+                }
+            }));
+            test.sendData(node, 'input', "isTransient?");
+        }).should.become("isTransient? false");
+    });
+    it("should have isTransient flag that can be set to true", function() {
```
dbooth-boston, line 243: good

src/framework-ondata.js
dbooth-boston: I think it would also be wise to modify deleteVni() in the vni-manager.js to delete the inputStates.  input-states.js does not seem to have a deleteAllInputStates method, and that module is a bit gnarly so it might be easiest to ask James to add that function, so that you can call it from deleteVni().

This change could be made after getting an initial version working, since it only helps with memory cleanup.  In fact, I'm not even sure it is absolutely needed.  But it seems like it would be wise, to help make sure that memory leaks don't crop up.

glenna-m: Let's discuss at stand up tomorrow

glenna-m: And yes, you did explicitly tell me to delete them all.  We had even discussed using that method which is why I coded that way.  You are changing the design from what you spec'd last week.

src/noflo-component-access.js
```diff
@@ -22,7 +22,8 @@ function facadeComponent(node) {
         get componentName() {
             return node.componentName;
         },
-        isTranslator: node.isTranslator
+        isTranslator: node.isTranslator,
+        isTransient: node.isTransient
```
jamesrdf: When would node.isTransient be anything but undefined?

src/noflo-component-factory.js
```diff
@@ -38,6 +38,7 @@ module.exports = function(nodeDef, callback){
             inPorts: _.mapObject(nodeDef.inPorts, changeMulti2Addressable)
         });
         node.isTranslator = _.isUndefined(nodeDef.isTranslator) ? false : nodeDef.isTranslator;
+        node.isTransient = _.isUndefined(nodeDef.isTransient) ? false : nodeDef.isTransient;
```
catch-point: Why not set the facade directly and leave the noflo node alone?

dbooth-boston:
LGTM

src/noflo-component-factory.js
glenna-m: I think I made the change you are requesting, which does make sense since both isTranslator and isTransient are pipeline flags, not core noflo.    Please take a look and let me know.

catch-point:
LGTM

PR #160, opened 2016-08-24
Marking the metadata manipulation components as transient

glenna-m:
Main change here is marking the metadata manipulation components as transient for the pipeline.  While I was doing this, I also added more user friendly descriptions and icons.


dbooth-boston:
LGTM

PR #161, opened 2016-08-24
Changes required to enable Ahlta pipeline.

glenna-m:
Changes include: 
- Recognize if metadata is already set for the VNI and use it; Ahtla path requires this because the "official" id is set up stream, while the data query demographic will have a different id in it.  Patient hash should not tamper with the official ID.
  - While in there, added a description and icon for better user experience.


dbooth-boston:
LGTM

PR #162, opened 2016-08-24
New id-hash-mapper component

glenna-m:
This component allows for mapping an id into a specific pool of ids, based on an input object attribute.
The attribute used is specified with a json pointer.  This component has a couple key benefits over
our original id-mapper component:
1. It does not require a one-to-one id map - that gets tedious with a million ids
2. When configured correction, it can ensure appropriate id mapping policies
       such as male ids mapped to male procedures and female ids mapped to
       female procedures for more reasonable data.


dbooth-boston:
LGTM

PR #164, opened 2016-08-25
And-gate component

glenna-m:
Used by the pipeline to ensure we received data from both CMUMPS and AHLTA hashes


dbooth-boston:
LGTM

catch-point:
LGTM

PR #166, opened 2016-08-25
Remove default cmumps authority IP from code

catch-point:
The noflo graph will require the authority to be set and the (skipped) test requires an environment variable to be set


dbooth-boston:
LGTM

glenna-m:
LGTM

PR #170, opened 2016-08-26
Must not be transient for pipeline or the IIP name and value ports ar…

glenna-m:
…e empty after first time.


dbooth-boston:
LGTM

PR #171, opened 2016-08-31
new write-content component

glenna-m:


dbooth-boston:
LGTM

This looks helpful for debugging.

catch-point:
Why not write to a file in the temporary directory, using os.tmpdir()? This way it could also be used in production environments (with read-only deployments).

glenna-m:
I think you are referring to the unit test use of /tmp?  The os.tmpdir() API does not seem to work correctly on the Mac.  I get errors when I use it and then try to access the file created - that is why I am using /tmp.  I could create a directory for this kind of output under the test directory, if you like that approach better.   The write-content component itself just writes to whatever file the user specified.

catch-point:
okay, LGTM then

PR #172, opened 2016-08-31
Making URN format consistent with ShEx URN format

glenna-m:
Modifications to support more consistent URN naming (which is already in use by the ShEx component:

urn:local:fhir:&lt;patientId if it exists&gt;:&lt;translator&gt;:&lt;resource id&gt;

### DIFF
--- a/components/cmumps2fhir.js
+++ b/components/cmumps2fhir.js
@@ -101,9 +101,9 @@ function graphUri(translation) {
 
         // TODO: Replace the hard coded patientID with a metadata key
         if (_.isUndefined(state.patientId)) {
-            uri = 'urn:local:' + encodeURIComponent(translator) + ':' + resourceId;
+            uri = 'urn:local:fhir::' + encodeURIComponent(translator) + ':' + resourceId;
         } else {
-            uri = 'urn:local:' + state.patientId + ':' + encodeURIComponent(translator) + ':' + resourceId;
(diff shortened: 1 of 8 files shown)


components/cmumps2fhir.js
```diff
@@ -101,9 +101,9 @@ function graphUri(translation) {
 
         // TODO: Replace the hard coded patientID with a metadata key
         if (_.isUndefined(state.patientId)) {
-            uri = 'urn:local:' + encodeURIComponent(translator) + ':' + resourceId;
+            uri = 'urn:local:fhir:' + encodeURIComponent(translator) + ':' + resourceId;
         } else {
-            uri = 'urn:local:' + state.patientId + ':' + encodeURIComponent(translator) + ':' + resourceId;
+            uri = 'urn:local:fhir:' + state.patientId + ':' + encodeURIComponent(translator) + ':' + resourceId;
```
dbooth-boston: 1. I didn't think about this before, but we should probably call encodeURIComponent on the patientId and resourceId too, for safety.
2. If there is no patientId, I would rather still have that field in the URN retained but empty, so that it can be consistently parsed, like urn:local::myTranslator:1234

glenna-m: OK.  I'll make both changes.

catch-point:
LGTM

dbooth-boston:
LGTM

PR #174, opened 2016-09-02
Fix/verify outstanding issues are no longer issues:

glenna-m:
#31: create-state.js should allow state creation with undefined lm
#33: Do not change output.lm when the output.error becomes true

### DIFF
--- a/src/framework-ondata.js
+++ b/src/framework-ondata.js
@@ -46,7 +46,7 @@ module.exports = function(payload, socketIndex) {
          // Get the old & new input state for port and the new payload
          var lastInputState = vni.inputStates(portName, socketIndex);
          var inputState = (_.isUndefined(payload.vnid)) ? 
-             createState(vnid, payload, undefined, undefined, undefined, undefined, nodeInstance.componentName) 
+             createState(vnid, payload, createLm(), undefined, undefined, undefined, nodeInstance.componentName) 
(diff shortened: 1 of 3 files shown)


src/create-state.js
```diff
@@ -50,7 +50,7 @@ module.exports = function(vnid, data, lm, error, stale, groupLm, componentName)
                 error: error,
                 stale: stale,
                 groupLm: groupLm,
-                lm: lm || createLm(),
+                lm: (arguments.length > 2) ? lm : createLm(),
```
dbooth-boston: Actually I think this was correct the way it was before, because lm should not be left undefined if data is defined.

src/framework-ondata.js
```diff
@@ -46,7 +46,7 @@ module.exports = function(payload, socketIndex) {
          // Get the old & new input state for port and the new payload
          var lastInputState = vni.inputStates(portName, socketIndex);
          var inputState = (_.isUndefined(payload.vnid)) ? 
-             createState(vnid, payload, undefined, undefined, undefined, undefined, nodeInstance.componentName) 
+             createState(vnid, payload, createLm(), undefined, undefined, undefined, nodeInstance.componentName)
```
dbooth-boston, line 49: good

dbooth-boston:
LGTM

PR #57, opened 2016-03-08
Tidying up code a bit.  No logic changes.

glenna-m:
Some minor clean up to our code.   No real logic changes here.

PR #175, opened 2016-09-08
Fix issue #57

glenna-m:
IIPs should not be cleared when transient node VNI is deleted

### DIFF
--- a/src/framework-ondata.js
+++ b/src/framework-ondata.js
@@ -308,7 +308,11 @@ function handleOutput(vni, port, lastLm, state ) {
             port.disconnect();
 
             if (nodeInstance.isTransient) { 
-                vni.delete();
+                if (vni.vnid === '') { 
+                    vni.clearTransientInputs();
+                } else {
+                    vni.delete();
+                }
             }
 
             return true;
--- a/test/framework-ondata-mocha.js
+++ b/test/framework-ondata-mocha.js
@@ -17,9 +17,11 @@ var noflo = require('noflo');
 var createLm = require('../src/create-lm');
 var framework_ondata = require('../src/framework-ondata');
 var factory = require('../src/pipeline-component-factory');
+var logger = require('../src/logger');
 var jswrapper = require('../src/javascript-wrapper');
 var stateFactory = require('../src/create-state');
-var logger = require('../src/logger');
+var vniManager = require('../src/vni-manager');
+
 var test = require('./common-test');
 
 describe("framework-ondata", function() {
@@ -65,7 +67,7 @@ describe("framework-ondata", function() {
                 
             });
 
-should.be.rejectedWith('No wrapper fRunUpdater function found!  Cannot run updater.');
+            should.be.rejectedWith('No wrapper fRunUpdater function found!  Cannot run updater.');
         });
 
         it("should manage single port node vni state, fRunUpdater invocation, & output state", function() {
@@ -469,8 +471,9 @@ should.be.rejectedWith('No wrapper fRunUpdater function found!  Cannot run updat
             // Define the fRunUpdater that framework should invoke
             var fRunUpdater = function(vni) { 
                 // Verify this is vni context
-                vni.should.have.all.keys(
-                    'vnid', 'delete', 'inputStates', 'errorState', 'outputState', 'nodeInstance');
+                vni.should.have.all.keys( 'vnid', 'clearTransientInputs', 
+                                          'delete', 'inputStates', 'errorState', 
+                                          'outputState', 'nodeInstance');
                 
                 test.verifyState(vni.inputStates('input1'), '', input1);
                 test.verifyState(vni.inputStates('input2'), '', input2);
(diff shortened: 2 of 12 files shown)


test/framework-ondata-mocha.js
```diff
@@ -1048,102 +1063,159 @@ should.be.rejectedWith('No wrapper fRunUpdater function found!  Cannot run updat
...
-                     console.log.restore();
-                     done.should.equal(testData);
+                     test.verifyState(done, '1', 'Buon pomeriggio! Arrivederci!');
 
                      // Verify VNI has been cleared
                      expect(nodeInstance).to.not.be.empty;
                      nodeInstance.vnis.should.be.an('object');
-                     nodeInstance.vnis.should.be.empty;
+
+                     var states = nodeInstance.vni('1').inputStates();
```
dbooth-boston: This looks like it is only checking to see if the input states.input port has been cleared.  Shouldn’t it check that the whole VNI 1 was deleted?

test/framework-ondata-mocha.js
```diff
@@ -1048,102 +1063,159 @@ should.be.rejectedWith('No wrapper fRunUpdater function found!  Cannot run updat
...
+                     // Verify we got the correct state which drew from both IIP and non IIP state data
+                     test.verifyState(done, '1', 'Salve! Come va? Addio!');
 
-                     // Verify default VNI has been cleared
-                     expect(nodeInstance).to.not.be.empty;
-                     nodeInstance.vnis.should.be.an('object');
-                     var vnids = Object.keys(nodeInstance.vnis);
-                     vnids.should.not.contain(''); // no default VNI
-                     vnids.should.have.length(5); // we should still have 5 other VNIs that were created
+                     // Now verify that the non IIP input state was cleared by the framework
```
dbooth-boston: I am confused here.  You might have to walk me through this.  If  am understanding correctly, we are on the targetNode at this point, and updater has fired under vnid 1 (but not vnid '').  If so, then shouldn't VNI 1 be deleted? And VNI '' should be unaffected, right?

test/vni-manager-mocha.js
```diff
@@ -194,6 +195,135 @@ describe("vni-manager", function() {
...
+                    var leNode = network.processes.leNode.component;
+                    network.graph.addEdge('inputNode', 'out', 'leNode', 'input');
+                    network.graph.addInitial(présenter, 'inputNode', 'in');
+
+                });
+            });
+
+            it("should not fail if socket is not initialized yet", function(done) {
+
+                var enchanté = "Enchanté de faire votre connaissance";
```
dbooth-boston: The French is fun to read, but I prefer that we avoid non-ASCII characters in the source code unless they are needed.

test/vni-manager-mocha.js
```diff
@@ -194,6 +195,135 @@ describe("vni-manager", function() {
...
+                             }
+                          },
+                          {fRunUpdater: function(vni) {
+
+                               // verify we have the correct state on the VNI input port
+                               var inputStates = vni.inputStates()['input'];
+                               inputStates.should.be.an('array');
+                               inputStates.should.have.length(2);
+
+                               // Since we have both IIP and edge, we should clear both
```
dbooth-boston, line 301: Add comment:
// We might want to change this policy in the future

test/framework-ondata-mocha.js
glenna-m: There's not a real easy way to do that directly since as soon as you do nodeInstance.vni('1') it will create it.  However, I did rework this based on this comment.  I deleted the inputState check and replaced with:

```
                 Object.keys(nodeInstance.vnis).should.be.empty;
```

Since in this case the VNI has been deleted, there should be no VNI keys.

test/framework-ondata-mocha.js
glenna-m: Yes, VNI 1 was deleted.  This should be verifying that VNI '' is unaffected.  I have updated this to be more clear.

test/vni-manager-mocha.js
glenna-m: OK.  I was curious how well it would handle the extended latin set, which is why I used it.   And it works quite well.  However, since it concerns you, I will change it.

test/framework-ondata-mocha.js
```diff
@@ -1048,102 +1063,157 @@ should.be.rejectedWith('No wrapper fRunUpdater function found!  Cannot run updat
...
                  }).then(function(done) {
-                     console.log.restore();
-                     done.should.equal(testData);
+                     test.verifyState(done, '1', 'Buon pomeriggio! Arrivederci!');
 
                      // Verify VNI has been cleared
                      expect(nodeInstance).to.not.be.empty;
                      nodeInstance.vnis.should.be.an('object');
-                     nodeInstance.vnis.should.be.empty;
+                     Object.keys(nodeInstance.vnis).should.be.empty;
```
dbooth-boston, line 1141: ok

dbooth-boston:
LGTM

src/framework-ondata.js
```diff
@@ -308,7 +308,11 @@ function handleOutput(vni, port, lastLm, state ) {
             port.disconnect();
 
             if (nodeInstance.isTransient) { 
-                vni.delete();
+                if (vni.vnid === '') { 
+                    vni.clearTransientInputs();
```
catch-point, line 312: This behaviour is very subtle and could use a comments like:
- Clear empty-string vni states, that are not single IIPs, when only empty-string vni states are used as input

PR #176, opened 2016-09-14
Enforce unique resource ids

glenna-m:
This PR enforces unique resource ids by inserting the patient id for the following id format: 
&lt;resource type&gt;:&lt;patient id&gt;:&lt;resource id&gt;
This will be applied to all of the ids for demographics, diagnoses, prescriptions, and procedures.

This checkin also removes no longer used ShEx/lab code in patient-hash, since ShEX is now using a different path through the pipeline.


dbooth-boston:
LGTM

catch-point:
LGTM

PR #177, opened 2016-09-15
Implementation of Transient with  global pipeline & node def flags.

glenna-m:
The component (node definition) transient flag has priority over global pipeline setting if both exist.


dbooth-boston (approved):
LGTM

catch-point:
LGTM

PR #180, opened 2016-09-20
Support clearTransientInputs on multi ports

glenna-m:


dbooth-boston (approved):
LGTM.  As we discussed last week, this will likely need to change once the architecture is better worked out regarding 'transient', but I think it makes sense to go ahead with this for the moment.

catch-point:
LGTM

PR #182, opened 2016-10-07
Increase timeouts on several mocha test

dbooth-boston:
Some tests still fail for me, but the failures appear to be for other reasons.


glenna-m:
LGTM

catch-point:
LGTM

PR #183, opened 2016-10-10
Added a check to warn the user if the required library file does not exist

glenna-m:
This checkin adds a check to tests that depend on the Java xml to RDF translation library to verify the library is installed and available in the standard location.  If the library is not there, the tests will fail with a confusing error message.  

This change ensures the user knows the reason for the test failure so they can locate and install the library and then run the test successfully.


dbooth-boston (approved):
LGTM

catch-point:
LGTM

PR #185, opened 2016-10-11
Sample basic memory test

glenna-m:
This is a simple sanity check  test on whether a component is growing the heap or consuming free memory after garbage collection.  I have some other tests, but would like to be sure we agree on this general approach prior to check in. 

You will notice I removed the references to the Winston logger in the funnel component and am using console.log.  I did this because I found that Winston logger is consuming significant additional memory on every call to it.

### DIFF
--- a/components/funnel.js
+++ b/components/funnel.js
@@ -2,9 +2,9 @@
 
 var _ = require('underscore');
 var fs = require('fs');
+var os = require('os');
 
 var stateFactory = require('../src/create-state.js');
-var logger = require('../src/logger.js');
 var profiler = require('../src/profiler.js');
 var wrapper = require('../src/javascript-wrapper.js');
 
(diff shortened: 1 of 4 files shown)


dbooth-boston (approved):
LGTM, but I noticed some console.log messages in funnel.js.  Were those left there intentionally, rather than going to the log?   Just wondering.

glenna-m:
As noted in the checkin notes, I pulled out the calls to the Winston logger from this component because it was accumulating significant memory - as much as 3 MB in processing 100 ids.  I do want the console messages on this component because they allow me to both monitor progress on the pipeline and to debug the pipeline when something goes wrong.  I suspect Winston is possibly one of the ongoing sources of our memory growth.

catch-point:
If we want to switch back to using console, I think we should be replacing logger.info with console.info. This will preserve the log level, which may be useful if we want to do anything else with the logging in the future.

PR #188, opened 2016-10-13
Use Custom console logger

catch-point:
Create a light facade around the console for logging and silencing the output

### DIFF
--- a/src/logger.js
+++ b/src/logger.js
@@ -1,39 +1,61 @@
 // logger.js
 
-var path = require('path');
 var _ = require('underscore');
-var winston = require('winston');
-var stackTrace = require('stack-trace');
+
+var levels = [
+    'debug',
+    'dir',
+    'log',
+    'info',
+    'warn',
+    'error',
+    'assert'
+];
 
 /**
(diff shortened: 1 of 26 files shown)


src/logger.js
```diff
@@ -1,39 +1,43 @@
...
-        })
-    ]
-});
+module.exports = {
+    time: console.time.bind(console),
+    timeEnd: console.timeEnd.bind(console),
+    silence: function(limit) {
+        levels.forEach(function(level) {
+            if (!limit || limit == 'all' || levels.indexOf(level) <= levels.indexOf(limit))
+                this[level] = _.noop;
```
dbooth-boston, line 24: Would it be helpful (i.e., easier for callers) to save the previous logging level (by pushing it onto a stack) when 'silence' is called, so that it can restored later (by popping it) without having to know what it was before?  Usage something like this:

```
silence();   // Temporarily quiet the logging
. . . Do noisy stuff here . . . 
unsilence();   // Restore the previous logging level
```

dbooth-boston:
LGTM

glenna-m:
LGTM

PR #191, opened 2016-10-27
3 memory tests enabled with --expose-gc

glenna-m:
- file-loader-mocha.js
- funnel-mocha
- xml-to-rdf-mocha.js

### DIFF
--- a/components/funnel.js
+++ b/components/funnel.js
@@ -1,9 +1,12 @@
 // funnel.js
 
 var _ = require('underscore');
+
 var fs = require('fs');
+var os = require('os');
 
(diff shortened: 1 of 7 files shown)


dbooth-boston (approved):
LGTM

src/logger.js
```diff
@@ -12,10 +12,15 @@ var levels = [
     'assert'
 ];
 
+var BYTES_TO_MB = 1024 * 1024;
+
 /**
  * Creates a custom logger for this package
  */
 module.exports = {
+    bytesToMb: function(num) {
```
catch-point: I think this would be better in the common-test file as it is more about formatting test results

PR #196, opened 2016-11-25
Frame resulting jsonld from shex-wrapper using vocab

catch-point:
Using a vocab, makes the shex jsonld output easier to read


dbooth-boston:
LGTM

glenna-m:
LGTM

PR #197, opened 2016-12-08
Issue 56, took CMUMPS authority out of source code

dbooth-boston:


glenna-m:
This will break the pipelines we currently have checked in.  You probably want to update those before merging this change down.

glenna-m:
Resolved with PR 215.  That PR fixed a bug in this PR with passing Patient ID to request-template, and included some test changes to run on Node 7.8.0

PR #198, opened 2016-12-15
Fix silent failures when RDF load fails

glenna-m:
Added a catch block to catch and correctly handle errors if RDF Store load fails


dbooth-boston (approved):
LGTM

PR #201, opened 2016-12-15
Enhanced file-loader to work as a split, or sequentially use default VNI

glenna-m:
The original file-loader was doing a variation on split - spawning one VNI per line, and flooding the network with all at once (and eating a lot of memory).
If the use_default_vni flag is set, this version will send one VNI at a time, using the default VNI, throttling the processing and eating less resources.


dbooth-boston (approved):
LGTM

PR #202, opened 2016-12-16
New component for sampling the attributes in a large dataset

glenna-m:


package.json
```diff
@@ -27,6 +27,7 @@
     "bluebird": "^3.4.0",
```
dbooth-boston, line 27: Does this need to use bluebird, or can it use standard promises?

glenna-m: This component does not use bluebird at all.  It uses standard promises.  The white on the diffs indicates something that was not changed around the green changes.  Bluebird is still in there because some of Eric's code, unrelated to this PR, still uses it.

dbooth-boston: ok

dbooth-boston:
LGTM

PR #204, opened 2017-02-02
Made replace-substring into a transient component

glenna-m:


dbooth-boston (approved):
LGTM

In general, I'd like to move away from nodes/components that just do little things like this, but if this is needed at the moment then it's fine.  I hope that we can eventually raise the functionality of the pipeline so that little things like this are not needed.  Let's talk when we can about how we could eventually get there.

glenna-m:
This component already existed.  My change here was to make it transient so I could ensure consistent URI names given the two inputs.  And I added tests since there were none.   It could be replaced when we refactor shex but right now, I need it.

PR #205, opened 2017-02-06
Better icon for demo purposes

glenna-m:
Changing from the default icon to a column comparison icon for better usability


dbooth-boston:
LGTM

PR #206, opened 2017-02-06
Improve write-content json formatting.

glenna-m:
Original  code wrote everything into a difficult to read string; this change ensures written JSON is properly formatted for easier reading and analysis.


dbooth-boston (approved):
LGTM

PR #208, opened 2017-02-08
New domain filter component

glenna-m:
Domain filter component for demo
Uses the new filters library filterByJsonPointer API


dbooth-boston (approved):
LGTM

PR #212, opened 2017-03-30
Jsonld-utils library.  First API is a getContext API.

glenna-m:
This API will be used to replace the hard coding in the original ShEx component and will be used in the new ShEx component.

### DIFF
--- a/components/lib/jsonld-utils.js
+++ b/components/lib/jsonld-utils.js
@@ -1,0 +1,66 @@
+/**
+ * jsonld-utils.js
+ *
+ * A library of JSON-LD helper APIs for use in the noflo RDF pipeline component updaters.
+ */
+
+var _ = require('underscore');
+
+var request = require('request');
+var url = require('url');
+
+var logger = require('../../src/logger');
+
+module.exports = {
+    getContext: getContext 
+};
+
+/** 
+ * Checks the JSON-LD context to see if it is a URL, or a fully described context.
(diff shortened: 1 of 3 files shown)


components/lib/jsonld-utils.js
```diff
@@ -0,0 +1,65 @@
...
+function getContext(data, defaultContext) {
+
+    if (_.isEmpty(data)) return Promise.reject("getContext called with no data!");
+
+    // Verify we have something that looks like JSON-LD - it should have @context and @graph.  
+    if (_.isEmpty(data["@context"]) && _.isUndefined(defaultContext)) return Promise.reject("getContext called with no context!");
+    if (_.isEmpty(data["@graph"])) return Promise.reject("getContext called with no data graph!");
+
+    return new Promise(function(resolve, reject) {
+        var context = _.isEmpty(data["@context"]) ? graphContext : data["@context"];
```
dmtaub: Where is graphContext defined?

dbooth-boston (approved):
LGTM

components/lib/jsonld-utils.js
glenna-m: It wasn't.  Good catch.  At one point it was defaultContext, but I had changed it and not cleaned it up.  Did not see it in testing since you couldn't get there without a context anymore.    I cleaned up and thought of another test to improve coverage.  Please take a look and let me know if you have other thoughts on it.

dmtaub:
LGTM

PR #218, opened 2017-05-12
Sequential file loader component used in round trip ShEx pipeline demo

glenna-m:
Also cleaned up a little test noise and tested an upgraded noflo-groups package


dbooth-boston (approved):
LGTM

